PR #1, opened 2019-09-26
Initial configuration

julienw:
In this PR I set up the basic stuff:
* linting (eslint and flow)
* build (babel)
* dev configuration (using nodemon and babel-node)
* CI
* git and install hooks

I copied a lot of stuff over from the profiler UI repository.
To check that linting / flow / babel work fine, I added some files that aren't included in this PR, but please tell me if you'd rather like that they're included instead.

Also the separate commits are meaningful so you can look at the PR looking at the individual commits if you prefer.


.babelrc
```diff
@@ -0,0 +1,16 @@
+{
+  presets: [
+    ["@babel/preset-env", { targets: { node: true }}],
```
julienw: This is different from the profiler where we target browsers using browserlist.

.babelrc
```diff
@@ -0,0 +1,16 @@
...
+    ["@babel/preset-flow", { all: true }]
+  ],
+  plugins: [
+    // Without `loose`, the transformation uses `Object.defineProperty` which
+    // obeys the spec more, but we get a bundle that's 8kB larger. That was the
+    // default in babel v6.
+    ["@babel/plugin-proposal-class-properties", { loose: true }],
+    ["@babel/plugin-proposal-object-rest-spread", { useBuiltIns: true }]
+  ]
+}
```
julienw: There's no "test" specific configuration anymore: we should have the same set-up now... although I haven't configured tests yet so maybe I'll have to bring it back.

.eslintrc.js
```diff
@@ -0,0 +1,77 @@
+module.exports = {
```
julienw: This is the same config, except I removed everything about react and jsx.

.flowconfig
```diff
@@ -0,0 +1,10 @@
+[options]
+suppress_comment=\\(.\\|\n\\)*\\$FlowExpectError
+suppress_comment=\\(.\\|\n\\)*\\$FlowFixMe
+include_warnings=true
+
+[ignore]
+.*node_modules/.*
+
+[libs]
+src/types/libdef
```
julienw: I removed the "globals" directory for now.

bin/post-checkout.js
```diff
@@ -0,0 +1,10 @@
+// @noflow
+
+const [origHead, head, flag] = process.env.HUSKY_GIT_PARAMS.split(' ');
```
julienw: I only removed the "compatiblity layer" for the older husky, because we don't have it here.

package.json
```diff
@@ -0,0 +1,57 @@
...
+    "flow-generate-libdefs": "flow-typed install --libdefDir src/types/libdef",
+    "lint": "eslint .",
+    "lint-fix": "yarn lint-js --fix",
+    "preinstall": "node bin/pre-install.js",
+    "build:clean": "rimraf dist && mkdirp dist",
+    "build": "babel src --out-dir dist",
+    "serve": "node dist/index.js",
+    "start": "nodemon --exec babel-node -w src src/index.js",
+    "start:prod": "run-s build:clean build serve",
+    "test-all": "# noop"
```
julienw: This does nothing but because it is used by the git hooks, I added it.

package.json
```diff
@@ -0,0 +1,57 @@
...
+    "flow-stop": "flow stop",
+    "flow-coverage": "flow-coverage-report -i '**/*.js' -t html -t text",
+    "flow-generate-libdefs": "flow-typed install --libdefDir src/types/libdef",
+    "lint": "eslint .",
+    "lint-fix": "yarn lint-js --fix",
+    "preinstall": "node bin/pre-install.js",
+    "build:clean": "rimraf dist && mkdirp dist",
+    "build": "babel src --out-dir dist",
+    "serve": "node dist/index.js",
+    "start": "nodemon --exec babel-node -w src src/index.js",
```
julienw: I tried for some time to make it work without babel-node but this is more reliable like this.

PR #2, opened 2019-09-26
Add flow-typed

julienw:
I forgot to add flow-typed in the previous PR.


src/types/libdef/npm/@babel/cli_vx.x.x.js
```diff
@@ -0,0 +1,90 @@
+// flow-typed signature: 0344b8033965c759fd85dad0ffa3655f
```
julienw: I added all the generated libdefs even if that's a bit useless, just so that it's easier to generate the next ones.

PR #4, opened 2019-09-26
Add a http framework

julienw:
This introduces [the framework koa](https://koajs.com/). Koa is made by the same people as express, but is more lightweight and embraces async functions. That's why I decided to use koa over express because of its more modern architecture. Tell me what you think!

(please do not look at the first commit as it is #2).


package.json
```diff
@@ -53,5 +54,10 @@
       "post-rewrite": "node bin/post-rewrite.js",
       "pre-push": "yarn test-all"
     }
+  },
+  "dependencies": {
+    "@koa/router": "^8.0.1",
+    "convict": "^5.1.0",
```
julienw: convict is used by firefox send to manage options. It's handy because it automatically fetches options from environment variables, with default values suitable for the dev environment.

src/routes.js
```diff
@@ -0,0 +1,14 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
julienw: For the routes I decided to use a simple file instead of a directory with individual files because we won't have many endpoints. But it's easy to change anyway :)

julienw: acutally, I take that back and will make a directory :)

package.json
gregtatum: Great! Does it give good error messages then if the environment variables are malformed? This seems to be a common source of bugs and deployment issues.

src/config.js
```diff
@@ -0,0 +1,24 @@
...
+
+const conf = convict({
+  httpPort: {
+    format: Number,
+    default: 4243,
+    env: 'HTTP_PORT',
+  },
+});
+
+conf.validate();
```
gregtatum: Ah, here is a validation.

src/config.js
```diff
@@ -0,0 +1,24 @@
...
+
+type Config = {|
+  +httpPort: number,
+|};
+
+const config: Config = conf.getProperties();
+
+export default config;
```
gregtatum: 👍

src/routes/index.js
```diff
@@ -0,0 +1,14 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+// @flow
+
+import Router from '@koa/router';
+
+export default function() {
+  const router = new Router();
+  router.get('/', ctx => {
+    ctx.body = 'Hello world!';
```
gregtatum: Oh hi 🌎

package.json
julienw: Yeah it should; but then we'll need to put stricter rules maybe, like custom validation functions. Currently a lot are just "string"...

src/config.js
julienw: This was copied from firefox send ;)

julienw: This was just copied from Firefox Send to be honest :)

PR #6, opened 2019-09-30
Slightly change the configuration object

julienw:


src/config.js
```diff
@@ -6,16 +6,25 @@
 import convict from 'convict';
 
 const conf = convict({
+  env: {
+    doc: 'The application environment.',
+    format: ['production', 'development', 'test'],
+    default: 'development',
+    env: 'NODE_ENV',
+  },
```
julienw: I added this, taking the code of firefox send as an example.
I don't use it yet, but this can come handy.

PR #7, opened 2019-09-30
Add dockerflow routes

julienw:
This PR implements some of the requirements from https://github.com/mozilla-services/Dockerflow.

I'm implementing tests for these endpoints in a separate PR.


bin/generate-version-file.js
```diff
@@ -0,0 +1,57 @@
...
+  if (!isAtGitRoot) {
+    throw new Error('This script must be run at the root of the repository.');
+  }
+}
+
+function writeVersionFile() {
+  const repositoryUrl = process.env.CIRCLE_REPOSITORY_URL;
+  const commitHash = process.env.CIRCLE_SHA1;
+  const buildUrl = process.env.CIRCLE_BUILD_URL;
+  const version = '1.0';
```
julienw: I considered fetching this from package.json, but then settled for the simpler path. Please tell me what you think!

gregtatum: I'm not really understanding what this version means exactly. Is this the version of our app? Is it of the container? When does this update?

Generally, I'd prefer to keep version numbers in more "declaration"-like places. I think you can use `require` on `json` files. `require('../package.json').version`.

src/routes/dockerflow.js
```diff
@@ -0,0 +1,58 @@
...
+// @flow
+
+// In this router we implement the services required by the dockerflow mozilla
+// best practices, as described in https://github.com/mozilla-services/Dockerflow.
+
+import Router from '@koa/router';
+import fs from 'fs';
+import path from 'path';
+
+export default function() {
```
gregtatum: What do you think about exporting named functions? It seems like they are easier to refactor since you can identify them across the codebase. e.g. `export function dockerFlowRoutes () {}`

src/routes/dockerflow.js
```diff
@@ -0,0 +1,58 @@
...
+    // easy to mock in tests (maybe impossible). Also path.join does no
+    // filesystem access so it's probably faster than them.
+    const fileName = path.join(__dirname, '../../dist/version.json');
+
+    try {
+      const content = await fs.promises.readFile(fileName);
+      ctx.body = content;
+      ctx.type = 'application/json';
+    } catch (e) {
+      if (e.code === 'ENOENT') {
```
gregtatum: It'd probably be good to document this code. I'm not sure what it is without looking it up.

`ENOENT`.

Also, I think if this is not found it should be a 500, right? 404 is if the end user mistyped something. Here we're internally not configured.

src/routes/dockerflow.js
```diff
@@ -0,0 +1,58 @@
...
+        ctx.status = 404;
+      } else {
+        ctx.status = 500;
```
gregtatum: How about also adding a nice error message here? That'll make it easier to debug in case of failures.

gregtatum: If we're doing JSON responses, maybe we can standardize on something like `{ error: "The version file could not be found." }`

src/routes/dockerflow.js
```diff
@@ -0,0 +1,58 @@
...
+  });
+
+  //
+  // "Respond to /__lbheartbeat__ with an HTTP 200. This is for load balancer
+  // checks and should not check backing services."
+  router.get('/__lbheartbeat__', ctx => {
+    ctx.body = 'OK';
+  });
+
+  return router;
```
gregtatum: 👍 looking good

bin/generate-version-file.js
julienw: I don't really know what the version means either... should it increase each time we deploy? Or is it just informational (and in that case we don't really have a use for it). That's why I kept it here for now.
I asked our contact at devops for some more context about this value.

src/routes/dockerflow.js
julienw: in this PR I'm doing it for this one only but I'll follow-up for the other cases in the codebase.

bin/generate-version-file.js
julienw: From the doc in https://github.com/mozilla-services/Dockerflow/blob/master/docs/version_object.md:

*  The release version of the application
   *  can be left blank for non-release versions

So I think I'll do that for now, and revisit when we'll need something better.

PR #8, opened 2019-10-01
Add a test framework with tests for the dockerflow endpoints

julienw:
In this PR we implement the tests for https://github.com/firefox-devtools/profiler-server/pull/7.
Please do not look at the 3 first commits as they're part of that PR. The other commits should be fairly easy to look separately.


src/types/libdef/npm/supertest_vx.x.x.js
```diff
@@ -0,0 +1,45 @@
+// flow-typed signature: cc2e46fdf76da83aa88e0bfaefb15345
```
julienw: It's a bit sad that there's no libdef for this library.

.circleci/config.yml
```diff
@@ -17,5 +17,6 @@ jobs:
           key: v1-dependencies-{{ checksum "package.json" }}-{{ checksum "yarn.lock" }}
       - run: yarn lint
       - run: yarn flow:ci
+      - run: yarn test --runInBand
```
julienw: I decided to not run the coverage (yet?).

Also I'm not comfortable fetching a bash script from a web page to execute in an environment that will have secrets (in the future)...

test/setup.js
```diff
@@ -0,0 +1,14 @@
...
+
+afterEach(() => {
+  // The configuration to restore and reset all of the mocks in tests does not seem
+  // to be working correctly. Manually trigger this call here to ensure we
+  // don't get intermittents from one test's mocks affecting another test's mocks.
+  //
+  // See https://github.com/facebook/jest/issues/7654
+  jest.resetAllMocks();
+  jest.clearAllTimers();
+});
```
julienw: copied over from the profiler UI

package.json
```diff
@@ -41,11 +42,13 @@
...
     "flow-typed": "^2.6.1",
     "husky": "^3.0.5",
+    "jest": "^24.9.0",
     "mkdirp": "^0.5.1",
     "nodemon": "^1.19.2",
     "npm-run-all": "^4.1.5",
     "prettier": "^1.18.2",
-    "rimraf": "^3.0.0"
+    "rimraf": "^3.0.0",
+    "supertest": "^4.0.2"
```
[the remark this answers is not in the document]
julienw: I tried chakram and this is a bit different, because it's really testing as a black box, which makes it impossible to mock anything. So I think `supertest` is the best solution among the ones I know.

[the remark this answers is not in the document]
gregtatum: I don't have opinions here. My last node project I spun up my own server in the tests and hit the end-points manually.

src/types/libdef/npm/supertest_vx.x.x.js
gregtatum: We should probably add some incrementally as we use it.

src/app.js
```diff
@@ -0,0 +1,20 @@
...
+
+import Koa from 'koa';
+
+import routes from './routes';
+
+function createApp() {
+  const app = new Koa();
+
+  const router = routes();
+  app.use(router.routes());
```
gregtatum: Optional: Maybe add some short comments here as to what's being configured for these `use` statements. That'll help new folks get up to speed faster.

test/api/dockerflow.test.js
```diff
@@ -0,0 +1,46 @@
...
+  it('answers to the version endpoint when the file is present', async () => {
+    const fixture = require('./fixtures/version.json');
+    jest.spyOn(fs.promises, 'readFile').mockResolvedValue(fixture);
+    const agent = setup();
+    const response = await agent
+      .get('/__version__')
+      .expect('Content-Type', /^application\/json/)
+      .expect(200);
+
+    expect(response.body).toEqual(fixture);
```
gregtatum: The superagent API seems pretty nice.

package.json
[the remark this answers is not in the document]
julienw: yeah that's what I did with chakram: see https://github.com/firefox-devtools/profiler-server/compare/master...julienw:use-chakram?expand=1#diff-a56d2ab82e2de422f75dd163066ac919R1.

But I'm not satisfied, and also it seems that because jest doesn't run afterEach when a test fails the test server isn't stopped in that case...

src/types/libdef/npm/supertest_vx.x.x.js
julienw: yeah good idea, I'll try to add some.

julienw: so, I ended up doing a lot of it, and contributed it back to the flow-typed project in https://github.com/flow-typed/flow-typed/pull/3587.

PR #9, opened 2019-10-18
Use workflows for CircleCI

julienw:


.circleci/config.yml
```diff
@@ -1,22 +1,66 @@
...
+
+workflows:
+  version: 2
+  master:
+    jobs:
+      - tests
+      - lint
+      - build
+      - flow
+      - generate-version-file
```
julienw: In the future, this one will be replaced by the docker build -- that includes the version file generation.

.circleci/config.yml
```diff
@@ -1,22 +1,66 @@
...
+      # With --frozen-lockfile, the installation will fail if the lockfile is
+      # outdated compared to package.json.
+      - run: yarn install --frozen-lockfile
       - save_cache:
           paths:
+            - .git
             - node_modules
           key: v1-dependencies-{{ checksum "package.json" }}-{{ checksum "yarn.lock" }}
+
+workflows:
```
julienw: Note: compared to how I did it in the profile UI, the workflows come before the jobs here. This has been suggested by a friend, and I thought it's indeed clearer this way. I think I'll eventually do it in the profile UI too.

PR #11, opened 2019-10-22
Use named exports only

julienw:
As we discussed previously :)


julienw:
FTR I tried this eslint rule on the profiler UI, and there were too many violations to handle them in a few minutes, so I didn't follow up. This is my WIP branch: https://github.com/julienw/perf.html/tree/no-default-export

PR #14, opened 2019-10-25
Add a request logger for development only

julienw:


package.json
```diff
@@ -44,6 +44,7 @@
     "flow-typed": "^2.6.1",
     "husky": "^3.0.5",
     "jest": "^24.9.0",
+    "koa-logger": "^3.2.1",
```
julienw: I added this in `devDependencies`, I think it's appropriate because we load in development mode only. But I'm not so sure :) tell me if you think otherwise!

gregtatum: I think it would be slightly safer to have it in the normal dependencies, but I'm happy to go with whichever way you want.

Risks in dependencies:
An unnecessary module is loaded in production.

Risks in devDependencies:
The logger mistakenly gets added to a production area somewhere in the future and crashes the production app.

julienw: I believe you're right, I'll change this then.

PR #24, opened 2019-11-22
Improve the logging mechanism

julienw:
Another super simple PR, that brings the following improvements, in 3 commits:
1. defines a Logger type. This will help keep a logger instance in variables and have it typed properly.
2. adjust the default logger level to "debug" instead of "verbose", but make it configurable to set it to verbose if necessary (including in production). This makes it possible to add super verbose logs without them showing up by default. The logLevel is also exposed so that we can change some behavior if necessary, and we use that in the following commit already.
3. the request logger isn't running in tests by default, but this change enables it when the log level is manually set to "verbose".

About 2: maybe it would be better to expose a function so that we could check for example `if (wouldBeLogged('info'))`. I didn't need that for now but tell me what you think. This could be useful for performance, so that if we want to compute something costly to log at "debug" we could avoid it. Tell me whart you think.

### DIFF
--- a/src/log.js
+++ b/src/log.js
@@ -7,13 +7,51 @@
 
 import mozlog from 'mozlog';
 
-import { config } from './config';
+type LowerCasedLogLevel =
+  | 'trace' //     This will gather a stack automatically.
+  | 'verbose' //   This is displayed only when run with LOG_LEVEL=verbose
+  | 'debug' //     This is displayed in test and development environments.
+  | 'info' //      This is displayed in production too.
+  | 'warn' //      This is for messages more important than info but that aren't errors.
+  | 'error' //     This is for non-fatal errors.
+  | 'critical'; // This is for fatal errors.
 
-const isProduction = config.env === 'production';
+function toValidLogLevel(logLevel: ?string) {
+  if (!logLevel) {
+    return null;
+  }
+
+  const lowerCasedLogLevel = logLevel.toLowerCase();
+  switch (lowerCasedLogLevel) {
+    case 'trace':
+    case 'verbose':
+    case 'debug':
+    case 'info':
+    case 'warn':
+    case 'error':
+    case 'critical':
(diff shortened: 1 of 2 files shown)


src/log.js
```diff
@@ -7,13 +7,54 @@
...
+    case 'critical':
+      return lowerCasedLogLevel;
+    default:
+      return null;
+  }
+}
+
+// Note we don't use `config` because we want `config` to be able to log using
+// this file, and we want to avoid cycles in dependencies.
+const isProduction = process.env.NODE_ENV === 'production';
```
julienw: This will be used in the next PR.

src/log.js
```diff
@@ -7,13 +7,54 @@
 
 import mozlog from 'mozlog';
 
-import { config } from './config';
+type LowerCasedLogLevel =
+  | 'trace'
```
gregtatum: Why are there so many levels here? I feel that this is overly complicated to use, and I wouldn't know which one to use. Can we get away with just a few tiers?

julienw: Ah, I just reused the ones from the underlying library.

I'd use:
* verbose for super verbose stuff
* debug for things we want to see during development
* info for things we want to see in production
* warn for recoverable errors (maybe ?)
* error for non-fatal errors
* critical for fatal errors

I can add comments explaining the levels.

`trace` seems to have a special handling, but I'm not sure really.

julienw: I wrote this:
```js
type LowerCasedLogLevel =
  | 'trace' //     This will gather a stack automatically.
  | 'verbose' //   This is displayed only when run with LOG_LEVEL=verbose
  | 'debug' //     This is displayed in test and development environments.
  | 'info' //      This is displayed in production too.
  | 'warn' //      This is for messages more important than info but that aren't errors.
  | 'error' //     This is for non-fatal errors.
  | 'critical'; // This is for fatal errors.
```

PR #26, opened 2019-11-22
Implement a mechanism to provide a configuration through a local file

julienw:
Such a mechanism will be useful once we start working with Google Storage, because we'll have to configure local values different from the production values.

The configuration values I'm talking about here are:
* the bucket name we use for development
* a filename containing the authentication key to access google's API endpoint.

An alternative is convention over configuration: the configuration for Google Storage will have hardcoded values. But I feel like this is less flexible.

**Important**: 
* there are 2 "flavors" for this PR, the last commit shows the difference. Please tell me your preference. I think my preference is when this last commit is present. The difference is how we test config.js.
* it's probably easier to read the last commit (and so the full PR) without whitespace changes.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,9 +15,6 @@
 # "slim" flavor because we don't need the big version.
 FROM node:10-buster-slim AS builder
 
-ENV NODE_ENV="production"
-ENV PORT=8000
-
 # Create the user we'll run the build commands with. Its home is configured to
 # be the directory /app. It helps avoiding warnings when running tests and
 # building the app later.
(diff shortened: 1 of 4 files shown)


julienw:
Removing the review request for now, there's a weird issue in tests with `require`, probably because I use `resetModules` in tests.

src/config.js
```diff
@@ -4,28 +4,55 @@
 // @flow
```
julienw: It's easier to see the changes in this file without whitespace changes.

Dockerfile
```diff
@@ -15,9 +15,6 @@
 # "slim" flavor because we don't need the big version.
 FROM node:10-buster-slim AS builder
 
-ENV NODE_ENV="production"
-ENV PORT=8000
-
```
julienw: When I did this initially I wasn't sure of it. Now I think this was a mistake. Especially this disturbs the tests.

Dockerfile
```diff
@@ -70,7 +67,7 @@ RUN set -x \
   && yarn test \
 # Actually build the project.
   && yarn build:clean \
-  && yarn build \
+  && NODE_ENV=production yarn build \
```
julienw: But I still want to run the build in production mode.

canova: Yeah, I think it's still good to have just in case

PR #29, opened 2019-12-09
Support a primitive form of versioning

julienw:
This implements a very simple form of versioning.
Once we deploy the server "for real" we'll want to implement an upgrade mechanism as well.

### DIFF
--- a/src/middlewares/versioning.js
+++ b/src/middlewares/versioning.js
@@ -6,11 +6,19 @@
 // This middleware ensures that the request specifies the versioning header and
 // its value is the expected value.
 // In the future we'll want to implement an upgrade mechanism.
+
 import type { Context } from 'koa';
 
-export const ACCEPT_VALUE_PREFIX =
-  'application/vnd.firefox-profiler+json;version=';
-export function versioning(version: string) {
+import contentType from 'content-type';
+
+// This mime-type is a vendor mime-type (because this starts with `vnd.`). It
+// also uses the 'json' suffix to outline that the content, if any, is in JSON form.
+// Lastly we specify the version as a mime-type parameter (;version=XXX).
+// Here is an acceptable value for the Accept header:
+//   application/vnd.firefox-profiler+json; version=1
+export const ACCEPT_VALUE_MIME = 'application/vnd.firefox-profiler+json';
+
+export function versioning(expectedVersion: number) {
   return async function(ctx: Context, next: () => Promise<void>) {
     const acceptValue = ctx.get('accept');
     if (!acceptValue) {
@@ -18,19 +26,24 @@ export function versioning(version: string) {
       return;
     }
 
-    if (!acceptValue.startsWith(ACCEPT_VALUE_PREFIX)) {
+    const parsedAcceptValue = contentType.parse(acceptValue);
+    if (parsedAcceptValue.type !== ACCEPT_VALUE_MIME) {
+      const expectedValue = contentType.format({
+        type: ACCEPT_VALUE_MIME,
+        parameters: { version: expectedVersion },
+      });
       ctx.throw(
         400,
-        `The header 'Accept' should have the value ${ACCEPT_VALUE_PREFIX}${version}`
+        `The header 'Accept' should have the value ${expectedValue}`
       );
       return;
     }
 
-    const requestedVersion = acceptValue.slice(ACCEPT_VALUE_PREFIX.length);
-    if (requestedVersion !== version) {
+    const receivedVersion = +parsedAcceptValue.parameters.version;
+    if (receivedVersion !== expectedVersion) {
       ctx.throw(
         406,
-        `Only the API version ${version} is supported by this server.`
+        `Only the API version ${expectedVersion} is supported by this server.`
       );
     }
 
--- a/src/middlewares/versioning.js
+++ b/src/middlewares/versioning.js
@@ -26,25 +26,32 @@ export function versioning(expectedVersion: number) {
       return;
     }
 
-    const parsedAcceptValue = contentType.parse(acceptValue);
-    if (parsedAcceptValue.type !== ACCEPT_VALUE_MIME) {
+    // An Accept header can contain several values separated by a comma.
+    // In this case this probably comes from a browser, and won't contain the
+    // value we're looking for here, but we're still looking at all values for
+    // correctness.
+
+    const acceptValues = acceptValue.split(',');
+
+    const hasAcceptableValue = acceptValues.some(acceptValue => {
+      const parsedAcceptValue = contentType.parse(acceptValue);
+      const receivedVersion = +parsedAcceptValue.parameters.version;
+      const isAcceptableValue =
+        parsedAcceptValue.type === ACCEPT_VALUE_MIME &&
+        receivedVersion === expectedVersion;
+      return isAcceptableValue;
+    });
+
+    if (!hasAcceptableValue) {
       const expectedValue = contentType.format({
         type: ACCEPT_VALUE_MIME,
         parameters: { version: expectedVersion },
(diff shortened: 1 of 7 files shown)


src/routes/publish.js
```diff
@@ -26,6 +27,7 @@ export function publishRoutes() {
 
   const router = new Router();
 
+  router.use(versioning('1.0'));
```
julienw: I decided to call `versioning` in this file, instead of a file higher in the route hierarchy, because when we'll want to implement upgrading we'll probably want to have a different mechanism for each API.

Open question: `1.0` or `1` ? I think I'd rather use `1` for simplicity, what do you think?

src/routes/index.js
```diff
@@ -6,10 +6,14 @@
...
 import { dockerFlowRoutes } from './dockerflow';
 import { publishRoutes } from './publish';
+import { versioning } from '../middlewares';
 
 export function routes() {
   const router = new Router();
   router.use(dockerFlowRoutes().routes());
+
+  // Versioning applies only to API routes, that's why we specify it here.
+  router.use(versioning('1.0'));
```
julienw: Initially I inserted `versioning` in the `publishRoutes` file, but I thought it was too misleading, because it wasn't applying only to that file.
To make that obvious I moved it here.

Open question: 1.0 or 1 ? I think I'd rather use 1 for simplicity, what do you think?

julienw: (I may find a better way to do it later)

src/middlewares/versioning.js
```diff
@@ -0,0 +1,39 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+// @flow
+
+// This middleware ensures that the request specifies the versioning header and
+// its value is the expected value.
+// In the future we'll want to implement an upgrade mechanism.
+import type { Context } from 'koa';
+
+export const ACCEPT_VALUE_PREFIX =
```
gregtatum: Please document this header value. It would be good to add some inline comments with the parts. I did a bit of research into why this was chosen, and it seems to be a `mime` type that specifies a specific version. It would be good to be a bit more explicit about why this approach was chosen.

test/api/publish.test.js
```diff
@@ -10,14 +10,19 @@ import { Storage as MockStorage } from '../../__mocks__/@google-cloud/storage';
...
 import { config } from '../../src/config';
 import { createApp } from '../../src/app';
+import { ACCEPT_VALUE_PREFIX } from '../../src/middlewares/versioning';
 
 beforeEach(() => MockStorage.cleanUp());
 afterEach(() => MockStorage.cleanUp());
 
 describe('publishing endpoints', () => {
   function setup() {
+    const acceptHeader = ACCEPT_VALUE_PREFIX + '1.0';
```
gregtatum: Should we have this line defined in `src/middlewares/versioning`?

test/api/publish.test.js
```diff
@@ -152,3 +157,29 @@ describe('publishing endpoints', () => {
...
+    // the log. Let's silence that.
+    jest.spyOn(process.stdout, 'write').mockImplementation(() => {});
+
+    const agent = supertest(createApp().callback());
+    return agent.post('/compressed-store').type('text');
+  }
+
+  it('returns an error when no `accept` header is specified', async () => {
+    const req = getPreconfiguredRequest();
+    await req.send('a').expect(400);
```
gregtatum: I think it would be beneficial to snapshot these message bodies, so we can tell if they change in the future.

test/api/publish.test.js
julienw: I thought of this too, but then I thought it was kind of nice to define it in 2 different ways in these 2 files -- so that if this changes we really have to come to the test and change it for real too.

src/types/libdef/npm-custom/content-type_v1.x.x.js
```diff
@@ -0,0 +1,23 @@
+// @flow
+declare module 'content-type' {
```
julienw: I've contributed this to the flow-typed project as well (including tests etc).

julienw:
Hey @gregtatum, I'm requesting a new review because I significantly changed the PR. You can look at the last commit only to see the changes, or look at the full PR to look at it again.
I think I handled all you requests in the process!

julienw:
@gregtatum while testing, I just realized that the code is not working properly, because `Accept` headers can have a list of values. Especially the code in `versioning` throws when requesting from a browser...
I'll fix this monday then.

julienw:
hey @gregtatum, I believe this is ready to look at it again. I added 2 commits on top of the previous one. Thanks!

src/middlewares/versioning.js
```diff
@@ -0,0 +1,59 @@
...
+    // correctness.
+
+    const acceptValues = acceptValue.split(',');
+
+    const hasAcceptableValue = acceptValues.some(acceptValue => {
+      const parsedAcceptValue = contentType.parse(acceptValue);
+      const receivedVersion = +parsedAcceptValue.parameters.version;
+      const isAcceptableValue =
+        parsedAcceptValue.type === ACCEPT_VALUE_MIME &&
+        receivedVersion === expectedVersion;
```
gregtatum: I guess in the future this will need to be modified to support version upgrading.

test/api/publish.test.js
```diff
@@ -152,3 +157,77 @@ describe('publishing endpoints', () => {
     );
   });
 });
+
+describe('API versioning', () => {
+  function getPreconfiguredRequest() {
+    // In these tests we'll generate errors, which triggers a lot of output from
+    // the log. Let's silence that.
+    jest.spyOn(process.stdout, 'write').mockImplementation(() => {});
```
gregtatum: Should we snapshot this output? In the future if this breaks, it might be confusing if we have no log information.

test/api/publish.test.js
```diff
@@ -152,3 +157,77 @@ describe('publishing endpoints', () => {
...
+    await req.send('a').expect(400, `The header 'Accept' is missing.`);
+  });
+
+  it('returns an error when an invalid `accept` header is specified', async () => {
+    const req = getPreconfiguredRequest().accept('INVALID_VALUE');
+    await req
+      .send('a')
+      .expect(
+        406,
+        `The header 'Accept' should have the value application/vnd.firefox-profiler+json; version=1`
```
gregtatum: 👍 nice test!

test/api/publish.test.js
```diff
@@ -152,3 +157,77 @@ describe('publishing endpoints', () => {
...
+    );
+    await req
+      .send('a')
+      .expect(
+        406,
+        `The header 'Accept' should have the value application/vnd.firefox-profiler+json; version=1`
+      );
+  });
+
+  it('accepts the request when there is the expected value amont several values', async () => {
```
gregtatum:
```suggestion
  it('accepts the request when there is the expected value among several values', async () => {
```

src/middlewares/versioning.js
julienw: Yeah, I think so. I have some ideas, but we'll probably want to register upgraders per API, so we'll need to think that a bit.

test/api/publish.test.js
julienw: Ah good idea, I'll look into that.

julienw: Ah good idea, I'll look into that.

julienw:
actually, snapshotting the errors made the tests fail, because errors have local paths :( I think I could merge because I was on a page with a previous github status...

julienw:
I'll fix this after my holidays, now it's too late :)

PR #30, opened 2019-12-09
Add security headers

### LINKED ISSUE #21: Add security headers
* add CORS
* add HSTS
* add CSP

I'd suggest to use https://github.com/venables/koa-helmet like Firefox Send does (see https://github.com/mozilla/send/blob/master/server/routes/index.js), because it seems more maintained than alternatives.

julienw:
Fixes #21.

Please don't look at the 2 first commits as they're #29. Otherwise you can look at the individual commits, I think they're meaningful and that should be easier to review.

I checked that I was properly getting a CSP report by adding this in `src/routes/index.js` (towards the top of the function):
```js
  router.get('/', async ctx => {
    ctx.body = '<div style='color: blue'>hello world</div';
  });
```

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -34,7 +34,8 @@ export function createApp() {
     reportTo([
       {
         group: 'cspreport',
-        maxAge: 365 * 24 * 60 * 60,
+        maxAge: 365 * 24 * 60 * 60, // 1 year
+        // The URL here should be identical to the URL we specify in the CSP policy below.
         endpoints: [{ url: '/__cspreport__' }],
       },
     ])
@@ -42,16 +43,20 @@ export function createApp() {
   app.use(
     helmet({
       contentSecurityPolicy: {
+        // because this is an API server and shouldn't be used as a webpage,
+        // everything is locked down as much as possible, following the
+        // checklist at https://github.com/mozilla-services/websec-check.
         directives: {
           defaultSrc: ["'none'"],
           frameAncestors: ["'none'"],
           baseUri: ["'none'"],
+          // This URI is what the checklist (see link above) suggests.
           reportUri: '/__cspreport__',
           reportTo: 'cspreport',
         },
       },
       hsts: {
-        maxAge: 31536000, // 1year
+        maxAge: 365 * 24 * 60 * 60, // 1 year
       },
     })
   );
--- a/src/app.js
+++ b/src/app.js
@@ -4,10 +4,12 @@
 // @flow
 
 import Koa from 'koa';
+import helmet from 'koa-helmet';
 
 import { config } from './config';
 import { getLogger, logLevel } from './log';
 import { routes } from './routes';
+import { reportTo } from './middlewares';
 
 const log = getLogger('app');
 
@@ -27,6 +29,32 @@ export function createApp() {
     // the server frontend instead.
     app.use(require('koa-logger')());
   }
+
+  app.use(
+    reportTo([
+      {
+        group: 'cspreport',
+        maxAge: 365 * 24 * 60 * 60,
+        endpoints: [{ url: '/__cspreport__' }],
+      },
+    ])
+  );
+  app.use(
+    helmet({
+      contentSecurityPolicy: {
+        directives: {
+          defaultSrc: ["'none'"],
+          frameAncestors: ["'none'"],
+          baseUri: ["'none'"],
+          reportUri: '/__cspreport__',
+          reportTo: 'cspreport',
+        },
+      },
+      hsts: {
+        maxAge: 31536000, // 1year
+      },
+    })
+  );
 
   // Adding the main endpoints for this app.
   // koa-router exposes 2 middlewares:
--- a/src/middlewares/index.js
+++ b/src/middlewares/index.js
@@ -6,3 +6,4 @@
 // This file just reexports all middlewares for an easier access.
 
 export * from './versioning';
+export * from './reportTo';
--- a/src/middlewares/reportTo.js
+++ b/src/middlewares/reportTo.js
@@ -1,0 +1,53 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+// @flow
+
+// This middleware adds the header Report-To that's used in CSP. This header
+// supersedes report-uri but isn't widely implemented nowadays.
+// To read more about it, here is the spec:
+// https://w3c.github.io/reporting/
+
+import type { Context } from 'koa';
+
+type Endpoint = {|
+  +url: string,
+  +priority?: number,
+  +weight?: number,
+|};
+
+type ReportToOptions = Array<{|
+  +group: string,
+  +maxAge: number,
+  +includeSubdomains?: boolean,
+  +endpoints: Array<Endpoint>,
+|}>;
+
+const HEADER_NAME = 'Report-To';
+
+export function reportTo(options: ReportToOptions) {
+  if (!options.length) {
+    throw new Error('reportTo requires at least one endpoint group.');
+  }
+
+  const readyOptions = options.map(
+    ({ group, maxAge, includeSubdomains, endpoints }) => ({
+      group,
+      max_age: maxAge,
+      include_subdomains: includeSubdomains,
+      endpoints,
+    })
+  );
+  // The header value is a the stringified JSON value without the start and
+  // stop brackets.
+  // See:
+  // - https://w3c.github.io/reporting/#header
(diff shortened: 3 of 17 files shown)


src/app.js
```diff
@@ -4,6 +4,7 @@
 // @flow
 
 import Koa from 'koa';
+import helmet from 'koa-helmet';
```
julienw: documentation for koa-helmet can be found at https://github.com/venables/koa-helmet

src/routes/publish.js
```diff
@@ -7,6 +7,7 @@
 // checks the data is uploaded to Google Cloud Storage.
 
 import Router from '@koa/router';
+import cors from '@koa/cors';
```
julienw: documentation for this is at https://github.com/koajs/cors

src/routes/publish.js
```diff
@@ -28,6 +29,7 @@ export function publishRoutes() {
   const router = new Router();
 
   router.use(versioning('1.0'));
+  router.use(cors());
```
julienw: I decied to use cors only for the API, so that we can't access other endpoints from web pages.

src/types/libdef/npm-custom/@koa/cors_v3.x.x.js
```diff
@@ -0,0 +1,19 @@
+// @flow
+
+declare module '@koa/cors' {
```
julienw: I contributed this back to flow-typed

src/types/libdef/npm-custom/koa-helmet_v5.x.x.js
```diff
@@ -0,0 +1,153 @@
+// koa-helmet explicitely supports the same API than helmet, so I imported the
+// libdef for helmet from https://github.com/flow-typed/flow-typed/blob/master/definitions/npm/helmet_v3.x.x/flow_v0.93.x-v0.103.x/helmet_v3.x.x.js
+// Then reformatted using prettier and improved for our usage.
```
julienw: Note: I didn't contribute this back to flow-typed because I didn't want to spend the time to write comprehensive tests for this... Please tell me if you'd prefer that I do it.

julienw:
Actually it's not ready for review yet, I have to implement the cspreport endpoint. Not sure what I should do with it though...

julienw:
This is now ready for review, thanks!

src/middlewares/index.js
```diff
@@ -0,0 +1,9 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+// @flow
+
+// This file just reexports all middlewares for an easier access.
+
+export * from './versioning';
```
gregtatum: I kind of want to do this pattern on the profiler front-end... The only real downside is that it broke jump to definition for me previously (not sure if it works now or not).

src/app.js
```diff
@@ -28,6 +30,32 @@ export function createApp() {
...
+        directives: {
+          defaultSrc: ["'none'"],
+          frameAncestors: ["'none'"],
+          baseUri: ["'none'"],
+          reportUri: '/__cspreport__',
+          reportTo: 'cspreport',
+        },
+      },
+      hsts: {
+        maxAge: 31536000, // 1year
```
gregtatum: I'd prefer this number to be `365 * 24 * 60 * 60 // 1 year`, like above, that way its consistent and not magical.

src/app.js
```diff
@@ -28,6 +30,32 @@ export function createApp() {
...
+    ])
+  );
+  app.use(
+    helmet({
+      contentSecurityPolicy: {
+        directives: {
+          defaultSrc: ["'none'"],
+          frameAncestors: ["'none'"],
+          baseUri: ["'none'"],
+          reportUri: '/__cspreport__',
```
gregtatum: What is this `__cspreport__`? It it part of the Mozilla recommendations? Perhaps it would be good to link to some docs on it.

src/app.js
```diff
@@ -28,6 +30,32 @@ export function createApp() {
...
+        maxAge: 365 * 24 * 60 * 60,
+        endpoints: [{ url: '/__cspreport__' }],
+      },
+    ])
+  );
+  app.use(
+    helmet({
+      contentSecurityPolicy: {
+        directives: {
+          defaultSrc: ["'none'"],
```
gregtatum: I feel like this section could use a few comments explaining why the values were chosen. I think it could be as simple as something like "this is an API and shouldn't be loaded as a webpage."

src/middlewares/reportTo.js
```diff
@@ -0,0 +1,53 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
gregtatum: I don't really understand this report-to mechanism.

Edit: Ok, things make more sense after reading the route implementation, perhaps it would be good to mention the route, and what happens to these reports once they get sent there.

src/routes/index.js
```diff
@@ -12,8 +14,9 @@ export function routes() {
   const router = new Router();
   router.use(dockerFlowRoutes().routes());
 
-  // Versioning applies only to API routes, that's why we specify it here.
+  // Versioning and CORS applies only to API routes, that's why we specify it here.
   router.use(versioning(1));
+  router.use(cors());
```
gregtatum: It might be worth mentioning that the CORS is using the default configuration here.

src/routes/cspReport.js
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    const cspReport = ctx.request.body['csp-report'];
+    if (!cspReport) {
+      // Send a "Bad Request" error as well if it doesn't look like a valid CSP report.
+      throw new BadRequestError(`The request misses a 'csp-report' field!`);
+    }
+
+    // Otherwise, just log the request
+    log.info('csp.report', cspReport);
```
gregtatum: This is a general observation and thought, no action is needed here:

I find I don't know what the various log levels are for, or where they get sent.

For instance, I'm assuming that this gets stored in server logs, but I'm not really sure. I think it might be worth simplifying this mechanism. Logs are important for local debugging, and server debugging, but they are also a source of potential PII leakage. It would be nice to have more confidence on what's going on when a log function is run, and what type of information gets stored there.

src/routes/cspReport.js
```diff
@@ -0,0 +1,36 @@
...
+
+// In this router we implement the service to handle csp reports.
+
+import Router from '@koa/router';
+import body from 'koa-json-body';
+
+import { getLogger } from '../log';
+import { BadRequestError } from '../utils/errors';
+
+export function cspReportRoutes() {
```
gregtatum: We should probably add some documentation for each route. It's a fairly "public" API, and would probably useful in the long term.

src/middlewares/index.js
julienw: ah, I don't know :) my own "jump to definition" is grep...

src/app.js
julienw: yeah it is... look at the "web applications" part of https://github.com/firefox-devtools/profiler-server/issues/18. I'll ask if there's another reference elsewhere.

src/app.js
julienw: yeah I can do that; actually I just used the CSP proposed in https://github.com/firefox-devtools/profiler-server/issues/18 :)

src/routes/cspReport.js
julienw: To be honest I'm not sure myself completely :)
I'm following the mozilla dockerflow recommendations, and they are that `info` logs are output in the console. That's all I know and I really don't know how they're handled later on... But I can find out!
Also I don't know how we act on these CSP reports later, the websec document doesn't say much about that, I'll also try to find out.

src/routes/cspReport.js
julienw: There's actually a comment above in the same file. Do you think it should be here instead?

src/app.js
julienw: yeah I can do that; actually I just used the CSP proposed in #18 :)

src/middlewares/reportTo.js
julienw: I added a few more words about that.

src/app.js
julienw: I added a link to https://github.com/mozilla-services/websec-check

src/app.js
julienw: I added a few words with a reference to the checklist too.

src/routes/cspReport.js
julienw: I moved the comment.

src/routes/cspReport.js
julienw: I filed https://github.com/firefox-devtools/profiler-server/issues/35 so that I don't forget about finding out.

PR #36, opened 2019-12-18
Remove snapshots because stacks sadly contain local paths

julienw:
In the last patch I added snapshots for the output logs. Sadly the errors contain local paths in the stack, and they're different on CI and local environments.
In this patch I remove the snapshots so that master stays green. I'll try to find a better solution when coming back from holidays.


julienw:
Thanks!

PR #37, opened 2019-12-24
Also watch the file .env to restart the server

julienw:
I had this PR ready for some time but I had a problem with nodemon (#1640) where that wouldn't work. Now with latest version I can't reproduce the issue anymore :/

It would be nice to test this on MacOS X because nodemon's implementation is vastly different on this system.


julienw:
Thanks for the review!

PR #40, opened 2020-01-14
Support CORS preflight requests

julienw:
CORS preflight requests are sent by the browser in some condition, that are met in the case of the profiler. Especially for us it's because we use a `upload` progress event handler.
A preflight request is a HTTP request using the `OPTIONS` verb and headers `Origin` and `Access-Control-Request-Method`, to the same path as the planned request.

These preflight requests are normally handled by the middleware `koa-cors` automatically when it's set up as expected. This wasn't the case in the previous code, and the result is that the middleware wasn't running for unknown routes (which was the case here because the `OPTIONS` verb matched nothing). This is a behavior from koa-router (https://github.com/koajs/router/issues/44).

The solution I used is attach the middleware directly on the app object. Koa always runs all middleware because it has no concept of route anyway.

### DIFF
--- a/src/routes/index.js
+++ b/src/routes/index.js
@@ -3,7 +3,6 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 // @flow
 
-import Router from '@koa/router';
 import cors from '@koa/cors';
 
 import { dockerFlowRoutes } from './dockerflow';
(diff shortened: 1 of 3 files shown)


test/api/publish.test.js
```diff
@@ -168,6 +168,20 @@ describe('publishing endpoints', () => {
...
+    const agent = supertest(createApp().callback());
+    let req = agent
+      .options('/compressed-store')
+      .set('Origin', corsHeaderValue)
+      .set('Access-Control-Request-Method', 'POST')
+      .send();
+
+    req = checkCorsHeader(req, corsHeaderValue);
+    await req;
+  });
```
julienw: I took care to write a failing test before fixing the problem.

gregtatum: Great, this helps on the review!

src/routes/index.js
```diff
@@ -12,16 +11,51 @@ import { cspReportRoutes } from './cspReport';
...
+ * preflight requests. So we have to specify the CORS middleware directly on the
+ * app object.
+ *
+ * For each router koa-router exposes 2 middlewares:
+ * - the result of routes() returns the configured routes.
+ * - the result of allowedMethods() configures the OPTIONS verb.
+ * We need to add them both if we want that the OPTIONS verb works properly
+ * (besides CORS).
+ */
+export function configureRoutes(app: Koa) {
```
gregtatum: 👍 Really easy to read. Thanks for the good comments explaining the design here.

PR #41, opened 2020-01-25
Return a JWT token after uploading a profile

julienw:
In this PR, instead of returning the profile token (aka the profile hash) we return a JWT token containing this profile token. In the future this JWT token will be used when the user who uploaded the profile will want to delete the profile.

This is probably better to review commit by commit.


tools/decode_jwt_payload.py
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/env python3
```
julienw: hey @JohanLorenzo, by chance would you have the bandwidth to review this simple python script with the corresponding test for me? Thanks a lot :-)

julienw: nevermind @johanlorenzo, our nice contributor @brisad offered to help on the review!

JohanLorenzo: 👍 

I'm sorry, I hadn't had the time to look at it with the current all-hands.

gregtatum: I'm not sure of the "why" of this python script.

src/logic/jwt.js
```diff
@@ -0,0 +1,25 @@
...
+  +profileToken: string,
+|};
+
+class JwtConfigurationError extends Error {
+  name = 'JwtConfigurationError';
+  status = 500;
+  expose = true; // The message will be exposed to users
+}
+
+export function getToken(payload: Payload): string {
```
gregtatum: Could this be named `generateToken` or `createToken`, I was a bit confused that it was actually doing the signing process. Or maybe also include Jwt in the title, since that's how it's used in `publish`

test/api/publish.test.js
```diff
@@ -19,6 +20,13 @@ import {
 beforeEach(() => MockStorage.cleanUp());
 afterEach(() => MockStorage.cleanUp());
 
+function verifyAndDecodeJwtToken(res) {
+  const token = res.text;
+  const secret = config.jwtSecret;
+  const decodedPayload = jwt.verify(token, secret, { algorithms: ['HS256'] });
```
gregtatum: I wonder if we can redefine the type definition to require an algorithms field. Also, what was your decision process for HS256?

src/logic/jwt.js
```diff
@@ -0,0 +1,25 @@
...
+  if (config.jwtSecret) {
+    return jwt.sign(payload, config.jwtSecret);
```
gregtatum: Please add the algorithm here in the options so it's explicit. It would also be good to document the decision around the encoding algorithm.

src/logic/jwt.js
gregtatum: It would be nice to write a small justification for how and why we are using JWT here. Plus it could be helpful to link to some of the literature we reviewed:

https://auth0.com/learn/json-web-tokens/
https://www.chosenplaintext.ca/2015/03/31/jwt-algorithm-confusion.html
https://auth0.com/blog/a-look-at-the-latest-draft-for-jwt-bcp/

gregtatum: Also to be clear, I would save more detailed documentation for a future PR.

tools/decode_jwt_payload.py
julienw: The goal for this python script is to support the use case of "upload a profile then gets the hash", in tests for example. We've seen it was advocated in some tests in the past. That's why I provide this python script, where we'd just need to pipe the output of the request to get the hash back.
I want to write some more detailed documentation later where I'd explain how one can use the API.

test/api/publish.test.js
julienw: 1. This is the default.
2. I didn't see any security reason to not use it. As long as we use a good secret it's secure.
3. We don't need to use a public/private key mechanism for our use of JWT tokens, so HS256 makes sense.

To be completely honest I don't exactly comprehend in which cases something like HS384 or HS512 would be necessary.

src/logic/jwt.js
julienw: ~~encoding~~ signing algorithm ;-)

src/config.js
```diff
@@ -50,6 +57,7 @@ type Config = {|
   +httpPort: number,
   +gcsBucket: string,
   +googleAuthenticationFilePath: string,
+  +jwtSecret: string,
```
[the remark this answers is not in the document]
julienw: Fixed this concern by using `null` as default value for this config, which makes it throw at start time if we don't set it using an environment variable.

test/api/publish.test.js
julienw: > I wonder if we can redefine the type definition to require an algorithms field.

TBH I was disappointed that the library isn't more opinionated about that nowadays.
To redefine the type definition I believe I'd have to copy it and I don't really want to do that. Instead I'll take care to encapsulate the logic in one place only.

PR #43, opened 2020-02-20
Update the CircleCI config to push the docker image to dockerhub

julienw:


.circleci/config.yml
```diff
@@ -4,6 +4,8 @@ executors:
   node:
     docker:
       - image: circleci/node:10.15
+    environment:
+      IMAGE_NAME: mozilla/firefoxprofiler
```
edunham: this should be `mozilla/profilerserver`

edunham: this should be `mozilla/profilerserver` as discussed on slack

.circleci/config.yml
```diff
@@ -82,3 +92,38 @@ jobs:
...
+      - setup_remote_docker
+      - run:
+          name: Load archived Docker image
+          command: docker load -i /tmp/workspace/image.tar
+      - run:
+          name: Publish Docker Image to Docker Hub
+          command: |
+            # Use both a version tag depending on the circle's build number
+            # and a latest tag that's depending on the branch name.
+            IMAGE_VERSION_TAG="0.0.${CIRCLE_BUILD_NUM}"
```
edunham: this will work for now but please plan to follow https://semver.org/ in prod

gregtatum: How does semver work in this situation? What would trigger a major/minor/patch change?

julienw: Yeah, we need to discuss this further. In my view we will never trigger major automatically and only will trigger minor. If we want to trigger major we could patch the variable here directly for example.

That said using the Circle build num isn't ideal because this won't increase version tags by 1, because there are builds for PRs too, not only for master.
So an alternative would be to use tags instead of merging to a production branch. Github makes it easy to tag from its interface, adding a changelog etc.

PR #50, opened 2020-03-23
Implement a shortening mechanism in the server

### LINKED ISSUE #46: Implement an endpoint for bitly shortening
We want to leverage the server to provide an endpoint for bitly shortening so that we can keep our access token secret.

julienw:
This implements a bitly client from the server. The goal is to hide the bitly access token.
To avoid malicious use of the bitly API through our server we allow only profiler URLs to be shortened through this API.

This was tested with the client side code in https://github.com/firefox-devtools/profiler/pull/2471.

It's probably easier to look at commits individually.

Fixes #46


package.json
```diff
@@ -78,6 +79,7 @@
     "koa-helmet": "^5.2.0",
     "koa-json-body": "^5.3.0",
     "koa-logger": "^3.2.1",
-    "mozlog": "^2.2.0"
+    "mozlog": "^2.2.0",
+    "node-fetch": "^2.6.0"
```
julienw: This utility has no dependency and makes the code easier to write and read.

package.json
```diff
@@ -51,6 +51,7 @@
     "husky": "^3.0.5",
     "jest": "^24.9.0",
     "mkdirp": "^0.5.1",
+    "nock": "^12.0.3",
```
julienw: This seems to be the most widely used tool to mock network requests in the node environment. It has very few dependencies as far as I've seen.

src/logic/shorten-url.js
```diff
@@ -0,0 +1,78 @@
...
+    method: 'POST',
+    headers: {
+      Authorization: `Bearer ${config.bitlyToken}`,
+      'Content-Type': 'application/json',
+    },
+    body: JSON.stringify(payload),
+  });
+
+  if (!response.ok) {
+    throw new Error(
```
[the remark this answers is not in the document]
gregtatum: Also, it might be good to try and read the error message, and log it or report it. I could see an error class do something like `new BitlyErrorResponse(response)`, then it could attempt to read the error message a bit more in-depth:

image

https://dev.bitly.com/v4/#operation/createBitlink

src/logic/shorten-url.js
```diff
@@ -0,0 +1,78 @@
...
+  if (!config.bitlyToken) {
+    log.error('No access token for bitly has been configured!');
+    throw new BitlyConfigurationError(
+      'No access token for bitly has been configured!'
+    );
+  }
+
+  const bitlyQueryUrl = 'https://api-ssl.bitly.com/v4/expand';
+  const payload = {
+    bitlink_id: urlToExpand.replace(/^https:\/\//, ''),
```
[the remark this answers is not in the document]
gregtatum: It's also not clear to me from the docs what is a valid string here: https://dev.bitly.com/v4/#operation/expandBitlink Do we need to remove the https? Can this be abused by providing domains that we don't support? Should we add an allow-list of domains? Is that over-engineering?

src/logic/shorten-url.js
```diff
@@ -0,0 +1,78 @@
...
+  const response = await fetch(bitlyQueryUrl, {
+    method: 'POST',
+    headers: {
+      Authorization: `Bearer ${config.bitlyToken}`,
+      'Content-Type': 'application/json',
+    },
+    body: JSON.stringify(payload),
+  });
+
+  if (!response.ok) {
```
gregtatum: Ditto on the custom error.

src/logic/shorten-url.js
```diff
@@ -0,0 +1,78 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// @flow
+import fetch from 'node-fetch';
+
+import { config } from '../config';
+import { getLogger } from '../log';
+
+class BitlyConfigurationError extends Error {
```
gregtatum: Should we collect all the errors in `utils/errors.js`?

src/routes/shorten.js
```diff
@@ -0,0 +1,77 @@
...
+    if (!longUrl.startsWith('https://profiler.firefox.com/')) {
+      throw new BadRequestError(
+        `Only profiler URLs are allowed by this service.`
+      );
+    }
+
+    const shortUrl = await shortenUrl(longUrl);
+
+    // Eslint thinks that ctx.body is assigned a value that depends on a
+    // previous value of ctx.body, which could be unsafe when using `await`. But
```
gregtatum: Should we disable this rule globally?

src/routes/shorten.js
```diff
@@ -0,0 +1,77 @@
...
+    const { shortUrl } = ctx.request.body;
+    if (!shortUrl) {
+      throw new BadRequestError(`The property 'shortUrl' is missing.`);
+    }
+
+    const longUrl = await expandUrl(shortUrl);
+
+    // The backend call has been made already, but still we want to discourage
+    // malicious users from using this API to expand any URL.
+    if (!longUrl.startsWith('https://profiler.firefox.com/')) {
```
gregtatum: Nice.

src/types/libdef/npm-custom/nock_v12.x.x.js
```diff
@@ -0,0 +1,143 @@
...
+  intercept(
+    path: $npm$nock$Path,
+    verb: string,
+    parameter?: $npm$nock$Parameter,
+    options?: any
+  ): this;
+  times(number: number): this;
+  once(): this;
+  twice(): this;
+  thrice(): this;
```
gregtatum: lol @ the method name

package.json
gregtatum: Thanks for thinking through the cost of extra dependencies, and limiting the "node_modules" bloat.

src/config.js
```diff
@@ -47,7 +47,7 @@ function loadConfig() {
     bitlyToken: {
       doc: `This token is used to access the bitly API`,
       format: String,
-      default: '',
+      default: process.env.NODE_ENV === 'test' ? 'token' : '',
```
gregtatum: A string like `'FAKE_TOKEN_FOR_TESTS'` may be a bit more explicit.

test/api/shorten.test.js
```diff
@@ -0,0 +1,226 @@
...
+    // An error has been output to the console.
+    expect(process.stdout.write).toHaveBeenCalledWith(
+      expect.stringContaining('server_error')
+    );
+
+    // There's been no request to the underlying service.
+    expect(nockScope.isDone()).toBe(false);
+  });
+
+  it('rejects with an error if the request does not have the expected properties', async () => {
```
gregtatum: Optional: If implementing a Bitly API error class like I suggested, then, test for the error response:

```
{
  "message": "string",
  "errors": [
    {
      "field": "string",
      "message": "string",
      "error_code": "string"
    }
  ],
  "resource": "string",
  "description": "string"
}
```

src/logic/shorten-url.js
julienw: I'm thinking of utils/errors to gather errors that are global to the program. Here it's clearly an error that's specific to this file.

I took this approach from the Result/Error approach in Rust, where each code unit implements errors specific to their business code.

src/logic/shorten-url.js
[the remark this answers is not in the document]
julienw: This was found by trial and error, the v3 API was easier for this :/

bitlink_id is actually returned by the 'shorten' API.

My approach here is that because we return an error if it's not a profiler URL, abusing this isn't very useful.
Also, we have to support bit.ly because some of our URLs use it :(

src/logic/shorten-url.js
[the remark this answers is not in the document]
julienw: yeah, good idea, I'll dig deeper.

src/routes/shorten.js
julienw: I don't know, it could be useful... but maybe too buggy for us?
I think I'd like to keep it for now, but we can disable it if it proves too painful.

src/config.js
julienw: good idea

Issue #28, opened 2019-12-03
Implement a profile deletion API


julienw:
We want to be able to delete an uploaded profile.

This is how I plan to implement this:
1. This depends on #22 so that a JWT is can be stored in the frontend.
2. This JWT identifies the user who published the profile. The JWT is passed using a header `Authorization: Bearer <token>`. This user is then authorized to delete the profile.
3. When the user is authorized we delete the profile in GCS.

The API could use the DELETE verb, for example:
`DELETE /v1/profiles/<token>`

This is the UI work: https://github.com/firefox-devtools/profiler/issues/2080.
See also [this Google document](https://docs.google.com/document/d/1HxNLWcz8qEZdBBLmRllSk7aBkwhBXnGQPbdOd_YTKtI/edit#heading=h.rxpj78ldlp7d).

julienw:
Done in #52

PR #52, opened 2020-03-25
Route to delete a profile

gregtatum:
Edit: this is now ready.

I still need to add tests, but this is the initial bit of work.

### DIFF
--- a/src/routes/index.js
+++ b/src/routes/index.js
@@ -8,6 +8,7 @@ import cors from '@koa/cors';
 import { dockerFlowRoutes } from './dockerflow';
 import { publishRoutes } from './publish';
 import { cspReportRoutes } from './cspReport';
+import { profileRoutes } from './profile';
 
 import { versioning } from '../middlewares';
 
@@ -50,9 +51,11 @@ export function configureRoutes(app: Koa) {
 function configureTechnicalRoutes(app: Koa) {
   const dockerFlow = dockerFlowRoutes();
   const cspReport = cspReportRoutes();
+  const profile = profileRoutes();
 
   app.use(dockerFlow.routes()).use(dockerFlow.allowedMethods());
   app.use(cspReport.routes()).use(cspReport.allowedMethods());
+  app.use(profile.routes()).use(profile.allowedMethods());
 }
 
 function configureUserFacingRoutes(app: Koa) {
--- a/src/routes/index.js
+++ b/src/routes/index.js
@@ -51,14 +51,15 @@ export function configureRoutes(app: Koa) {
 function configureTechnicalRoutes(app: Koa) {
   const dockerFlow = dockerFlowRoutes();
   const cspReport = cspReportRoutes();
-  const profile = profileRoutes();
 
   app.use(dockerFlow.routes()).use(dockerFlow.allowedMethods());
   app.use(cspReport.routes()).use(cspReport.allowedMethods());
-  app.use(profile.routes()).use(profile.allowedMethods());
 }
 
 function configureUserFacingRoutes(app: Koa) {
   const publish = publishRoutes();
+  const profile = profileRoutes();
+
   app.use(publish.routes()).use(publish.allowedMethods());
+  app.use(profile.routes()).use(profile.allowedMethods());
 }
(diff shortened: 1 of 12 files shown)


julienw:
Thanks for the work, I didn't check closely but I do have a few comments about the general architecture.
Tell me what you think!

1. Very commonly JWT token are sent using the header `Authorization`, using `Authorization: Bearer <token>`. Is there a reason you don't use this way of checking the token?

2. Doing it this way makes it very easy to use a middleware like https://github.com/koajs/jwt. Have you considered it? I think this could be used like this:
```
router.post(
  '/delete-profile',
   jwt({ secret, algorithms: ['HS256'], key: 'jwtData' }),
   async ctx => {
    // ctx.state.jwtData contains the decoded token
   });
```

I think that this would automatically return 401 in case the JWT is invalid. You could then also return 400 in case the token doesn't contain the expected data (the `profileToken`).

I believe this makes it very easy to implement and divide the responsibilities better.

gregtatum:
I'm not sure this is a general Authorization for a route. This seems more like a specific request using a token, which seems more like a `POST` content rather than a header. I think the `Authorization` field is more for general authentication for accounts for a given service. I'm not too opinionated here, so if that's how you want to do it, I can change the implementation.

gregtatum:
@julienw I added a reply to your comments, happy to change the implementation if you feel that I should switch it to headers. Please just "request changes" if that's the direction you want me to go.

gregtatum:
I almost wonder if we shouldn't change it to be

```
DELETE profile/{HASH}
Authorization: Bearer <token>
```

I think this would make the most semantic sense.

gregtatum:
Oops, I meant to break out the commit "Turn off require-atomic-updates"

gregtatum:
I'm going to re-work this to do: https://github.com/firefox-devtools/profiler-server/pull/52#issuecomment-604622133

julienw:
> I'm going to re-work this to do: [#52 (comment)](https://github.com/firefox-devtools/profiler-server/pull/52#issuecomment-604622133)

Good, this works for me!
This was also my first idea (see #28) but then thought that the information was redundant. But I agree this makes more sense in terms of API.
Thanks!

gregtatum:
Ok, this is now re-worked

src/routes/index.js
```diff
@@ -50,9 +51,11 @@ export function configureRoutes(app: Koa) {
 function configureTechnicalRoutes(app: Koa) {
   const dockerFlow = dockerFlowRoutes();
   const cspReport = cspReportRoutes();
+  const profile = profileRoutes();
 
   app.use(dockerFlow.routes()).use(dockerFlow.allowedMethods());
   app.use(cspReport.routes()).use(cspReport.allowedMethods());
+  app.use(profile.routes()).use(profile.allowedMethods());
```
julienw: This probably has to go to `configureUserFacingRoutes` below, otherwise it won't have CORS nor versioning.

(should I have added comments to these functions ? ;-) )

test/api/profile.test.js
```diff
@@ -0,0 +1,98 @@
...
+
+    await agent
+      .delete(`/profile/${profileToken}`)
+      .accept(acceptHeader)
+      .set('Authorization', `Bearer ${jwtToken}`)
+      .send()
+      .expect(400)
+      .expect('That profile was most likely already deleted.');
+  });
+});
```
julienw: It would be nice to add a test about CORS, see for example what I did for the bitly patch: https://github.com/firefox-devtools/profiler-server/pull/50/commits/2957888721d6383daab9e54d5a6128876aa400d8#diff-468c685187ffe91351f1c4acdde5516bR103-R117

As you'll see in that test, there are simple utilities to check for these headers.

test/api/profile.test.js
```diff
@@ -0,0 +1,98 @@
...
+      .expect('Profile successfully deleted.');
+
+    // Sending it another time will result in a 400 for it not existing.
+    await agent
+      .delete(`/profile/${profileToken}`)
+      .accept(acceptHeader)
+      .set('Authorization', `Bearer ${jwtToken}`)
+      .send()
+      .expect(400)
+      .expect('That profile was most likely already deleted.');
```
julienw: optional: because there's another test for the "not found" case, I believe that here you're mostly testing the mock (testing it returns 404).
Maybe checking `MockStorage.buckets[XXX].files[XXX]` exists before, and doesn't exist after, would be better.

PR #55, opened 2020-03-31
Unclutter the test output by silencing server errors for /profile

julienw:
This is a simple patch to make the test output easier to read :-)


julienw:
> Also, it's odd to me that we're logging server_errors for user errors.

I see what you mean. Currently we do this:
```
  app.on('error', err => {
    log.error('server_error', { error: err, stack: err.stack });
  });
```
I believe this is triggered when an error is thrown and caught by the framework. I guess we could refine this more.

PR #56, opened 2020-04-06
Upgrade dependencies

julienw:
This upgrades all dependencies in the codebase. I did it because github was reporting security issues.

I did some manual testing of our storage code (because the google lib was updated).


bin/post-rewrite.js
```diff
@@ -13,9 +13,9 @@ const rl = createInterface({
   input: process.stdin,
 });
 
-rl.on('line', line => {
+rl.on('line', (line) => {
```
gregtatum, line 16: I can't imagine the flame wars that lead to this change.

julienw: This is a configuration option, which means a lot for prettier :D
I kept the default because their ratonale makes sense (it's easier to add a type or add more parameters when the parentheses are already there).

More controversial (to me at least) is the decision to transform `function()` into `function ()` :D

julienw:
Thanks for the quick review!

PR #58, opened 2020-04-16
Add a performance stress test for the server

julienw:
This pull request introduces load tests for our server, backed by the tool [molotov](https://github.com/loads/molotov) written by our very own Tarek Ziadé.

This adds 2 tests: one for the publishing and deleting endpoints, and one for the url shortening and expanding endoints.

I wondered if we should run the scripts in CI but decided that it's not worth it. The rationale is that they're useful at one specific point in time (we won't run them often) so I believe it's reasonable that we let them diverge a bit from our endpoints, and update them when we need them.


loadtest/publish.py
```diff
@@ -0,0 +1,193 @@
...
+        # The environment always wins over what's defined in the file.
+        _API = endpoint_from_env
+
+    if _API is None:
+        raise ValueError(
+                'Please define one API endpoint before running the test file, '
+                'either uncommenting the right line in the file or specifying '
+                'the env variable API_ENDPOINT.'
+                )
+    print(f"We'll run tests against the endpoint {_API}.")
```
julienw, line 46: I wondered if I should share this in a common module. Do you think this would be useful? It would make it necessary to expose the API from the module instead of using the global like we do. I thought it would be a bit too complex for these 2 files, but maybe it would be nice to have a ready framework for the future... Tell me what you think!

loadtest/publish.py
```diff
@@ -0,0 +1,193 @@
...
+    pass
+
+
+@global_teardown()
+def test_ends():
+    """ This functions is called when everything is done.
+
+    Notice that it's not a coroutine.
+    """
+    pass
```
julienw, line 92: I wondered if I should keep them, because we don't use them. For now I kept them more as documentation of what we can do, in case we need it later.

gregtatum:
> I wondered if we should run the scripts in CI but decided that it's not worth it. The rationale is that they're useful at one specific point in time (we won't run them often) so I believe it's reasonable that we let them diverge a bit from our endpoints, and update them when we need them.

That's probably fine.

loadtest/publish.py
```diff
@@ -0,0 +1,193 @@
...
+    """
+    headers = dict(Accept='application/vnd.firefox-profiler+json;version=1')
+    return dict(headers=headers)
+
+
+@teardown()
+def worker_ends(worker_id):
+    """ This functions is called when the worker is done.
+
+    Notice that it's not a coroutine.
```
gregtatum, line 81: Why all the coroutine notices?

loadtest/publish_short_requests.py
```diff
@@ -0,0 +1,183 @@
...
+    """
+    _header, payload, _signature = jwt.strip().split(".")
+
+    decoded_str = jwt_base64_decode(payload)
+    json_payload = json.loads(decoded_str)
+    token = json_payload['profileToken']
+    return token
+
+
+def jwt_base64_decode(payload):
```
gregtatum, line 112: Optional: you should be able to collect these in a utility module.

loadtest/shorten.py
```diff
@@ -0,0 +1,119 @@
...
+        res = await resp.json()
+
+    long_url = res['longUrl']
+    return long_url
+
+
+# Each scenario has a weight. Molotov uses it to determine how often the
+# scenario is picked.
+# Here we have only one scenario so we don't use it.
+@scenario()
```
gregtatum, line 115: Optional: It might be nice to just pass in 1 for the scenario in case another gets added.

loadtest/README.md
```diff
@@ -0,0 +1,82 @@
...
+Please clean up afterwards :-)
+
+To run only one test:
+```
+molotov --max-runs 1 --single-mode <name of scenario> -v publish.py
+```
+Notice `-v` enables verbose mode, which outputs errors. This is especially
+useful while finalizing a specific scenario.
+
+To run the scenarii during 5 seconds, with 50 workers, this is how we can do it:
```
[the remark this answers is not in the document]
julienw: Ah, thanks! in french the plural is `scenarii` (can be `scenarios` too), I assumed it was the same in english.

loadtest/publish.py
julienw: This was autogenerated by the tool. I thought it was useful to keep the notice so that future developers would still get it right.

loadtest/publish_short_requests.py
julienw: Yeah, I wrote this second file very quickly by copy pasting... I believe you're right :-)

julienw: I collected them in a separate file, but I don't want to use it in `decode_jwt_payload.py` because I want that file to be standalone so that it can be copy pasted to use it elsewhere.

julienw: plot twist: this doesn't work with molotov right now (the module isn't found) if I don't transform all this in a package. This is more work than I want to do now. We'll do the work if we need this more later.

PR #59, opened 2020-04-17
Add flake8 to lint python

julienw:
Hey Greg, what do you think?

The drawback is that flake8 needs to be separately installed to be able to push. But the tool has been super useful to me while working on python code, to catch my errors before I tried to run scripts.


Dockerfile
```diff
@@ -31,7 +31,7 @@ RUN mkdir /app \
 # - python, as we need it to build fast-crc32c when installing with yarn later.
 # - build-essential, that installs all build tools for the same reason.
 RUN apt-get update \
-  && apt-get install -y --no-install-recommends git python build-essential \
+  && apt-get install -y --no-install-recommends git python build-essential flake8 \
```
gregtatum, line 34: I guess I never thought about this, but is our development docker environment the same as prod? Because this means that flake8 will be install on prod but never run, correct?

julienw: I'm glad you asked :-)

The Dockerfile is organized in 2 phases. Each phase start from a clean base image (see the command `FROM XXXX`).
Phase 1 is lint, tests and build. This is where flake8 is installed.
Phase 2 is the runtime container: we install just what's needed for runtime, and copy over what was built from Phase 1. So the runtime container is as light as needed.

I didn't invent this, this is the current best practices to build dockerfile. Firefox Send is doing this for example :-)

PR #62, opened 2020-05-04
Add a route for / so that it's easier to test using Mozilla observatory

julienw:
Mozilla observatory can only request /. On the server's current version, this errors. And because Koa unsets all headers on errors, all security headers are gone, and Mozilla observatory reports bogus results where we get some default CSP from our reverse proxies.

I think that Koa's behavior is unconditional: https://github.com/koajs/koa/blob/8d52105a34234be9e771ff3b76b43e4e30328943/lib/context.js#L132-L138
Therefore my "fix" here is to allow observatory observe a 200 response.
I would have liked to allow headers to be passed through but sadly this isn't possible. It's not a super big deal either.

### DIFF
--- a/src/routes/root.js
+++ b/src/routes/root.js
@@ -1,0 +1,32 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+// @flow
+
+// This router implements a very simple route for /, so that we can more easily
+// test the server through tools like the Mozilla Observatory.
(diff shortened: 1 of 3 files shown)


src/routes/root.js
```diff
@@ -0,0 +1,23 @@
...
+import { getLogger } from '../log';
+
+export function rootRoutes() {
+  const log = getLogger('routes.root');
+  const router = new Router();
+
+  router.get('/', async (ctx) => {
+    log.verbose('/');
+
+    ctx.body = 'OK';
```
gregtatum, line 19: Should we provide any additional information about what this page is to a user trying to access the API?

PR #65, opened 2020-05-07
Output more information to the version file

julienw:
When I wanted to know what version is deployed on which hostname I realized that the information from the `__version__` file isn't immediate to parse. There's the CircleCI build url that makes it possible to find information about when and what was built, but this requires to go to the build page and this is a slow operation.
Instead outputting the date and the branch will make it very immediate to know what we're dealing with.

I also output a `Last-Modified` header for this file. This should give roughly the same information than the `date` property so maybe this is redundant. Yet the information in the `date` property is a bit more immediate.

I wondered if I should embed this information in the version instead (eg: `0.0.circleci_build_number.datetime_in_YYYYMMDDHHMM_form.branch_name`) but decided this wasn't worth it. Indeed the version is computed both in the dockerfile and the node version script for simplicity, and I didn't want to give up this simplicity.

How the output file looks like:
image

The branch here isn't especially useful but I expect it will be `master` or `production` when built out of these branches -- the only ones that are pushed to dockerhub.

When checking if the server in docker is running (see especially the response headers):
Capture d’écran de 2020-05-07 12-32-04

You can look at the currently deployed `__version__`: [staging](https://dev.firefoxprofiler.nonprod.cloudops.mozgcp.net/__version__) (that one is a bit old) / [production](https://api.profiler.firefox.com/__version__) (it's currently a version from _master_ actually)

### DIFF
--- a/src/routes/dockerflow.js
+++ b/src/routes/dockerflow.js
@@ -27,6 +27,11 @@ export function dockerFlowRoutes() {
     // We try to get the version file in the current working directory.
     const versionFilePath = 'version.json';
     try {
+      const stat = await fs.promises.stat(versionFilePath);
+      ctx.lastModified = stat.mtime;
+
+      // We could try to conditionally answer depending on cache-related headers
+      // in the request but this is a lot of work for this small file.
       const content = await fs.promises.readFile(versionFilePath);
       ctx.body = content;
       ctx.type = 'json';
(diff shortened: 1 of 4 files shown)


src/routes/dockerflow.js
```diff
@@ -27,6 +27,11 @@ export function dockerFlowRoutes() {
     // We try to get the version file in the current working directory.
     const versionFilePath = 'version.json';
     try {
+      const stat = await fs.promises.stat(versionFilePath);
+      ctx.lastModified = stat.mtime;
+
+      // We could try to conditionally answer depending on cache-related headers
+      // in the request but this is a lot of work for this small file.
```
julienw, line 34: Also there are some ready-to-use middlewares to do all this, but I decided this wasn't worth it yet. Later if we need it to output more serious static resources we could revisit.

PR #67, opened 2020-05-07
Add some docs about deployment

julienw:
I wrote some explanations about our deployment situation and the servers we use. The goal is that anybody in the team should be able to find its way in the Google Cloud Platform maze.

We can augment this document when somebody doesn't find something :-)

Note: there's an embedded image that doesn't show up from the PR. To see it you need to choose "view file" from the context menu: 
image


docs-developer/deployment.md
```diff
@@ -0,0 +1,66 @@
...
+3. You need to select a project from the dropdown at the top.
+   1. Select from `firefox.gcp.mozilla.com` (dropdown at the top of the modal).
+   2. Then pick either `firefoxprofiler-prod` or `firefoxprofiler-nonprod`,
+      depending whether you want to look at the data for (respectively) the
+      production server or the staging server.
+4. Select `Logging` from the general menu, that you open with the top left
+   button.
+5. From the left dropdown, choose:
+   * for prod: Kubernetes Container > firefoxprofiler-prod-v1 > prod-firefoxprofiler > firefoxprofiler or nginx
+   * for staging: Kubernetes Container > firefoxprofiler-nonprod-v1 > dev-firefoxprofiler > firefoxprofiler or nginx
```
gregtatum, line 50: Yiiiikes. This is complicated.

Issue #68, opened 2020-05-11
Information about how to run the docker image doesn't work anymore


julienw:
There's some documentation in [docs-developer/docker.md](https://github.com/firefox-devtools/profiler-server/blob/master/docs-developer/docker.md) that doesn't seem to work anymore. Especially the environment variables don't seem to be picked up from `.env` like they should, and we need to provide information about how to provide the google configuration file, probably using `--mount` (see https://docs.docker.com/engine/reference/commandline/run/ for information about this command).

PR #69, opened 2020-05-11
Add a few docs about how to run the server

julienw:


julienw:
As a quick note, I noticed that the information about how to run the docker image doesn't work properly for external services, I filed #68 for this. I'll deal with it a bit later.

README.md
```diff
@@ -12,6 +12,67 @@ Here are some links to know more:
 
 The APIs are documented in [API.md](./API.md).
 
+## Develop and running the server locally
```
julienw, line 15: I wondered if that should be in eg `RUNNING.md` that would be linked from the README, instead of embedding everything in the README. Tell me what you think!

README.md
```diff
@@ -12,6 +12,67 @@ Here are some links to know more:
...
+### Install the local tooling
+You will need a recent enough version of [Yarn](http://yarnpkg.com/),
+version 1.10 is known to work correctly.
+You can install it into your home directory on Linux and probably OS X with:
+
+```bash
```
gregtatum, line 21: Maybe just link to: https://yarnpkg.com/getting-started/install

README.md
gregtatum: This is fine.

README.md
julienw: I just copied this from the profiler's README; but I'm not opposed.

julienw: So yarn's website has new instructions for yarn v2, this is a bit complex.
I added a line pointing to yarn v1's install insturctions.

PR #70, opened 2020-05-12
Fix docker instructions and support Google Cloud Storage authentication

### LINKED ISSUE #68: Information about how to run the docker image doesn't work anymore

julienw:
It may be a good idea to look at the commits separately, they're pretty much independent. But the second one is way bigger.

fixes #68


package.json
```diff
@@ -23,8 +23,8 @@
     "build": "babel src --out-dir dist",
     "generate-version-file": "node bin/generate-version-file.js",
     "docker:build": "docker build -t $npm_package_name:dev .",
-    "docker:run": "docker run -t -i -p 8000:8000 --rm --name $npm_package_name --env-file .env $npm_package_name:dev",
-    "docker:run:detached": "docker run -d -p 8000:8000 --rm --name $npm_package_name --env-file .env $npm_package_name:dev",
+    "docker:run": "docker run -t -i -p 8000:8000 --rm --name $npm_package_name --env-file .env \"$(node bin/output-docker-mount-option)\" $npm_package_name:dev",
+    "docker:run:detached": "docker run -d -p 8000:8000 --rm --name $npm_package_name --env-file .env \"$(node bin/output-docker-mount-option)\" $npm_package_name:dev",
```
julienw, line 27: I started doing a `docker-run.js` script, that would do all this. But I want to support the use case of running `yarn docker:run bash` for example, and this can be a bit difficult to add the new `--mount` command at the right location handling all possible cases (including these 2 commands attached/detached). So in the end I went this path of outputting just the command, which isn't so easy either but seems to work fine.

PR #73, opened 2020-05-13
Validate the configuration path for Google Cloud Storage at start time

julienw:
This is a simple patch to validate the configuration for GCS.
Indeed if the path is incorrect the GCS lib returns the underlying "file not found" error, which is [special-cased in koa](https://github.com/koajs/koa/issues/1420) to return a http status 404. When locally testing this was a bit painful.
Instead we validate this value when starting the server. I feel this is more correct.


src/config.js
```diff
@@ -54,7 +59,37 @@ function loadConfig() {
...
+  if (
+    googleAuthenticationFilePath &&
+    googleAuthenticationFilePath !== 'MOCKED'
+  ) {
+    // If some other value is set, we check it ahead of time so that the lib
+    // doesn't output an error later.
+    // Note that an empty string is a valid value: in that case the lib will try
+    // to find the authentication information from other means.
+    try {
+      fs.accessSync(googleAuthenticationFilePath, fs.constants.R_OK);
```
julienw, line 82: I use the synchronous version of the function because this is running at start time.

PR #74, opened 2020-05-13
Add warning mozlogs for authentication errors

julienw:
From the checklist in #18: `Access control failures must be logged at WARN level`.
So I add them for the "delete" route, that's the only one that uses an access control mechanism.


src/routes/profile.js
```diff
@@ -85,7 +102,7 @@ export function profileRoutes() {
       }
 
       ctx.body = 'Profile successfully deleted.';
-    }: Middleware)
```
julienw, line 88: I believe the cast wasn't needed really and this is a bit verbose.

src/routes/profile.js
```diff
@@ -52,19 +54,34 @@ export function profileRoutes() {
...
-          'The profileToken in the JWT did not match the token provided in the path.'
-        );
+        const message =
+          'The profileToken in the JWT did not match the token provided in the path.';
+        log.warn('delete.jwt.profileTokenMismatch', message);
+        throw new ForbiddenError(message);
       }
 
-      log.info(
+      log.debug(
```
julienw, line 84: I reduced the level for this log, I believe it's not really useful in production.

src/routes/profile.js
```diff
@@ -52,19 +54,34 @@ export function profileRoutes() {
         typeof jwtData !== 'object' ||
         typeof jwtData.profileToken !== 'string'
       ) {
-        throw new ForbiddenError(`A profileToken was not found in the JWT.`);
+        if (ctx.state.jwtOriginalError) {
+          const error = ctx.state.jwtOriginalError;
+          // The JWT decoding middleware found an error earlier.
+          log.warn(
+            'delete.jwt.invalid',
```
julienw, line 61: Note: I'll do a review of all the "codes" we use for logs, so that they're a bit more consistent accross the project. The goal will be to make it easier to extract important ones.

src/routes/profile.js
gregtatum: I think it was providing the type for the `ctx` variable, otherwise it was an any.

julienw: Ah right, I'll add the "Context" type in the parameters to get this. Thanks for the feedback!

julienw: plot twist: the Middleware type here coming from `@koa/router` types the context as `any` :)

julienw: So, I reused the same typing as you did before, I may have a look at it in a separate patch.

PR #78, opened 2020-05-18
Define and implement a convention for logs

julienw:
I tried to make some sense out of our logs and define a convention.
The goal is then to extract some of them to trigger alerts and draw graphs.

### DIFF
--- a/docs-developer/logging.md
+++ b/docs-developer/logging.md
@@ -1,0 +1,72 @@
+# Logging in the server
+
+We use [mozlog](https://github.com/mozilla/mozlog) for logging in the server.
+
+## Be careful with the API
+
+The API is a bit weird:
+* There are always 2 arguments.
+* The first argument must be a code without spaces. This is caught at runtime
+  when running in development, but because error cases aren't always tested it's
+  easy to miss, so extra care needs to be used.
+* The second argument can be anything, but we usually use a string or a json
+  object.
+
+## General usage
+
+First we get a logging object, then we log using a logging function specific to
+the loglevel we want to use.
+
+```js
+import { getLogger } from './log';
+const log = getLogger('code1');
+log.info('code2', message);
+```
+
+In the previous example, `code1` and `code2` are joined with a period to make up
+the `name` that's used in the logged message.
+
+## Convention to build names
+
+The names need to reflect where a log happens. In most cases we use a sequence
+containing:
+* the path to the file name, with periods as path separators;
+* the function where the log happens (can be skipped if there's only one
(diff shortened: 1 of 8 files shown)


test/api/profile.test.js
```diff
@@ -165,7 +165,7 @@ describe('DELETE /profile', () => {
       );
 
     expect(process.stdout.write).toHaveBeenCalledWith(
-      expect.stringMatching(/WARN.*delete.jwt.profileTokenMismatch/)
+      expect.stringMatching(/delete.*WARN.*jwt_profileTokenMismatch/)
```
julienw, line 168: This is a bit weird, and this is due to how the "pretty" formatter works currently. Digging into this more made me do https://github.com/mozilla/mozlog/pull/28 today, to make it more reasonable.

docs-developer/logging.md
```diff
@@ -0,0 +1,72 @@
...
+Here are the log levels we can use:
+* trace: this will gather a stack automatically. Note this is displayed only
+  when running the server with `LOG_LEVEL=trace`.
+* verbose: this is displayed only when runnint the server with `LOG_LEVEL=verbose` or `trace`.
+* debug: by default this is displayed in test and development environments.
+* info: this is displayed in production environment too.
+* warn: this is for messages more important than info but that aren't errors.
+  Especially access control errors need to use this level according to #18.
+* error: this is for non-fatal errors.
+* critical: this is for fatal errors.
```
gregtatum, line 68: I'm glad we have documentation for this, and I know that this is how other projects do logging, but this all feels horribly over-engineered for our use cases. It makes me a bit sad.

src/logic/shorten-url.js
```diff
@@ -39,9 +39,12 @@ class BitlyResponseError extends Error {
 }
 
 export async function shortenUrl(longUrl: string): Promise<string> {
-  const log = getLogger('shortenUrl');
+  const log = getLogger('logic.shorten_url.shorten_url');
```
gregtatum, line 42: This seems more reasonable `slug.slug_slug.slug` approach seems more reasonable given the restrictions.

docs-developer/logging.md
julienw: This could be slightly simplified but not so much actually. For the production server and getting alerts out of the logs I believe it's useful.

src/logic/shorten-url.js
julienw: I wasn't so happy for this specific case; maybe the file name should be changed to better reflect what's in it... bitly.js comes to my mind. #nitpicking :-)

PR #79, opened 2020-05-18
Remove trailing slash from the API endpoint if present

julienw:
I got bitten by this more than once, enough to add some code to support this :)

Basically before we had to take care that the env variable `API_ENDPOINT` doesn't have `/` at the end.

### DIFF
--- a/loadtest/publish.py
+++ b/loadtest/publish.py
@@ -43,6 +43,11 @@ def setup_api_endpoint():
                 'either uncommenting the right line in the file or specifying '
                 'the env variable API_ENDPOINT.'
                 )
+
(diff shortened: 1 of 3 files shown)


loadtest/publish.py
```diff
@@ -43,6 +43,11 @@ def setup_api_endpoint():
                 'either uncommenting the right line in the file or specifying '
                 'the env variable API_ENDPOINT.'
                 )
+
+    if _API[-1] == '/':
+        # Remove trailing slash if present.
+        _API = _API[:-1]
+
```
julienw, line 50: I know it would be better to have this in a library but I don't have time to do it now :(

PR #81, opened 2020-05-27
Increase the upload limit to 50MB

julienw:
This came while discussing with @fqueze that we could just increase this upload limit a bit for now. Given our current load and my load test I believe this is OK.


julienw:
Thanks!

PR #84, opened 2020-06-05
Implement a cheap payload "does that look like json" check

### LINKED ISSUE #61: Cheap checks on uploaded data
This is a quick stopgap for #20.

In this issue we want to do small checks that can be done while receiving the data.
The small check I have in mind is to make sure that the first character is a `{`.

julienw:
Fixes #61

The patch isn't so big but the stream API is confusing so I expect that not everything is straightforward. I tried to add comments to store my hard earned knowledge in the source files, for my future myself and coworkers, but please ask if you see anything that's unclear.

Also I split it in separate commits, with tests, so hopefully this should be easier to read by looking at the individual commits. (last commit is https://github.com/firefox-devtools/profiler-server/pull/85, please ignore it here)

I also tested uploading from the profiler UI, this was working properly.

### DIFF
--- a/src/routes/publish.js
+++ b/src/routes/publish.js
@@ -8,7 +8,7 @@
 
 import Router from '@koa/router';
 import Stream from 'stream';
-import util from 'util';
+import { promisify } from 'util';
 
 import { getLogger } from '../log';
 import { config } from '../config';
@@ -18,6 +18,9 @@ import {
   HasherPassThrough,
   LengthCheckerPassThrough,
   Concatenator,
+  CheapJsonChecker,
+  forwardErrors,
+  GunzipWrapper,
 } from '../utils/streams';
 import { PayloadTooLargeError } from '../utils/errors';
 
@@ -42,18 +45,67 @@ export function publishRoutes() {
 
     const hasherTransform = new HasherPassThrough();
     const lengthChecker = new LengthCheckerPassThrough(MAX_BODY_LENGTH);
+    // The payload should "look like" a json object.
+    const jsonChecker = new CheapJsonChecker();
+    const gunzipStream = new GunzipWrapper();
     const concatener = new Concatenator();
 
     // The "pipeline" utility conveniently destroys all streams when there's an
     // error. However the HTTP request isn't part of the pipeline because we
     // don't want to destroy it. Indeed if it's destroyed, the socket is closed
     // and we can't send a nice error to the caller.
-    const pipeline = util.promisify(Stream.pipeline);
+    const pipeline = promisify(Stream.pipeline);
+    const finished = promisify(Stream.finished);
+
+    // We create 2 interconnected pipelines:
+    //
+    //     request
+    //        |
+    //  lengthChecker --- gunzipStream --- jsonChecker
+    //        |
+    // hasherTransform
+    //        |
+    //   concatener
+    //
+    // These streams' possible errors are handled partly by nodejs' pipeline
+    // function, and some special error handlers to forward errors between
+    // pipelines.
+
+    // 1. First and main pipeline
     const pipelinePromise = pipeline(
       lengthChecker,
       hasherTransform,
       concatener
     );
+
+    // 2. "Gunzip and check" json pipeline
+    // This pipeline will stop quite early, once we know this looks like a json.
+    // When it ends we'll just unpipe.
+    // We don't use the `pipeline` utility because with it we get some
+    // "premature close" errors when the length checker errors out.
+    lengthChecker.pipe(gunzipStream);
+    gunzipStream.pipe(jsonChecker);
+    const jsonCheckerPromise = finished(jsonChecker);
+
+    // Stop the flow by unpiping when we're done checking the JSON.
+    jsonChecker.on('finish', () => {
+      log.verbose(
+        'json-checker-pipeline-done',
+        'The stream pipeline to check the json content is finished.'
+      );
+
+      // Note this is important to unpipe, because jsonChecker stopped consuming
+      // the data, and so if we don't stop the flow, the main pipeline will also
+      // stop flowing once the buffers in the second pipeline are full.
+      lengthChecker.unpipe(gunzipStream);
+      gunzipStream.destroy();
+    });
+
+    // Forward errors between streams in the second pipeline, as well as the
+    // first pipeline through lengthChecker.
+    // Note that errors inside the first pipeline are forwarded thanks to the
+    // `pipeline` thingie.
+    forwardErrors(lengthChecker, gunzipStream, jsonChecker);
 
     // The HTTP Request is piped directly to the first stream in the pipeline.
     // We do no error handling here: simply if there's an error (either expected
@@ -61,10 +113,10 @@ export function publishRoutes() {
     // anymore. The request's destroy should be properly handled by Koa or node.
     ctx.req.pipe(lengthChecker);
 
-    // We still wait for the end of the pipeline before moving forward.
+    // We still wait for the end of both pipelines before moving forward.
     // If there's an error (either expected or unexpected) it will bubble up to
     // Koa which will expose it appropriately to the caller.
-    await pipelinePromise;
+    await Promise.all([pipelinePromise, jsonCheckerPromise]);
 
     const storage = gcsStorageCreate(config);
     const hash = hasherTransform.sha1();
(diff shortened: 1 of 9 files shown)


src/routes/publish.js
```diff
@@ -42,29 +45,78 @@ export function publishRoutes() {
...
       lengthChecker,
       hasherTransform,
       concatener
     );
 
+    // 2. "Gunzip and check" json pipeline
+    // This pipeline will stop quite early, once we know this looks like a json.
+    // When it ends we'll just unpipe.
+    // We don't use the `pipeline` utility because with it we get some
+    // "premature close" errors when the length checker errors out.
```
julienw, line 85: Frankly I don't understand why we get these errors here but not in the previous pipeline... I spent too much time understanding all kind of errors and I'm a bit fed up with this, so I didn't want to invest more time now that everything works fine with the current implementation... I'm a bit frustrated that I don't understand this, so maybe I'll come back to it later, but I want to move on :-D

src/types/libdef/npm/jest_v25.x.x.js
```diff
@@ -901,7 +901,7 @@ type JestObjectType = {
 type JestSpyType = { calls: JestCallsType, ... };
 
 type JestDoneFn = {|
- (): void,
+ (error?: Error): void,
```
julienw, line 904: The change has been contributed to flow-typed: https://github.com/flow-typed/flow-typed/pull/3844 so there will be no problem updating the libdef later.

src/utils/streams.js
```diff
@@ -25,7 +28,7 @@ export class HasherPassThrough extends Transform {
 
   _transform(
     chunk: string | Buffer,
-    encoding: mixed,
+    encoding: string,
```
julienw, line 31: This was a mistake from previous patches.

src/utils/streams.js
```diff
@@ -36,7 +39,7 @@ export class HasherPassThrough extends Transform {
   // This method returns the computed sha1 from the data that passed through the
   // transform.
   sha1(): string {
-    this.log.verbose('sha1()');
+    this.log.trace('sha1()');
```
julienw, line 42: Using `trace` generates a stack automatically which is nice when debugging. To see trace logs we need to run tests with `LOG_LEVEL=trace`, they don't show up when the log level is "verbose", so it's also more silent.

src/utils/streams.js
```diff
@@ -110,18 +113,26 @@ export class Concatenator extends Writable {
...
+    this.log.trace('_destroy()');
     this.chunks.length = 0;
     this.contents = null;
-    callback();
+
+    // Passthrough the error information, if present.
+    // This line is needed because of the slightly inconsistent
+    // signature of callback vs err, and that we can't change.
+    err = err || undefined;
+    callback(err);
```
julienw, line 124: This was also a mistake from my side: we're supposed to forward the error to the callback. This API is so confusing and the documentation is missing some information...

src/utils/streams.js
```diff
@@ -130,3 +141,173 @@ export class Concatenator extends Writable {
...
+    }
+
+    // If we're coming here, this means we never finished checking. Let's
+    // happily throw, then!
+    callback(new BadRequestError(this.errorMessage));
+  }
+}
+
+// This simple wrapper simply encapsulates the native gunzip stream and rewrite
+// errors to make them more compatible with our code and koa's error handling.
```
julienw, line 236: So much code just to have nice errors :( I'm a bit sad.

src/utils/streams.js
```diff
@@ -279,3 +279,35 @@ export class GunzipWrapper extends Transform {
     callback(err);
   }
 }
+
+// This tool simply forward an error happening on a stream to other streams, by
+// destroying them.
+export function forwardErrors(
```
julienw, line 285: I didn't add unit tests for this utility but I believe the tests for the publishing cover it, by ensuring we get the right errors when feeding some bad payload to the server.

.circleci/config.yml
```diff
@@ -3,7 +3,7 @@ version: 2.1
 executors:
   node:
     docker:
-      - image: circleci/node:10.15
+      - image: cimg/node:10.21
```
julienw, line 6: This change has been extracted to https://github.com/firefox-devtools/profiler-server/pull/85, it's in this PR just so that the tests pass.

bin/pre-install.js
```diff
@@ -104,9 +104,7 @@ function parseExpectedNodeVersion() {
   const circleConfig = fs.readFileSync('.circleci/config.yml', {
     encoding: 'utf8',
   });
-  const expectedNodeVersion = /image: circleci\/node:([\d.]+)/.exec(
-    circleConfig
-  );
+  const expectedNodeVersion = /image: cimg\/node:([\d.]+)/.exec(circleConfig);
```
julienw, line 107: This change has been extracted to #85, it's in this PR just so that the tests pass.

test/api/publish.test.js
```diff
@@ -128,24 +130,18 @@ describe('publishing endpoints', () => {
...
   });
 
-  it('returns an error when the sent data is bigger than the length', async () => {
-    jest.spyOn(process.stdout, 'write').mockImplementation(() => {});
-    const req = getPreconfiguredRequest();
-    await req.set('Content-Length', String(3)).send('aaaa').expect(400); // 400 means Bad Request. It's generated automatically by Koa.
-    expect(process.stdout.write).toHaveBeenCalledWith(
-      expect.stringContaining('server_error')
-    );
-  });
```
julienw, line 138: I remove this test case because:
* I couldn't get the error I wanted with the changes
* this is auto-handled by node and koa -- when I tested, I got results like 400, 500 or 405 (method not allowed). The underlying node error is actually "invalid method" (see https://github.com/nodejs/node/issues/33728 for more explanation about that)
* this is quite an edge case so I didn't feel it was important to invest time to make a test for it after all.

test/unit/streams.test.js
```diff
@@ -0,0 +1,78 @@
...
+  it('accepts normal content', async () => {
+    const fixture = '{ "foo": "bar" }';
+    const checker = new CheapJsonChecker();
+    const input = new PassThrough();
+    input.write(fixture.slice(0, 3));
+    input.end(fixture.slice(3));
+    await expect(pipeline(input, checker)).resolves.toBe(undefined);
+  });
+
+  it('accepts content with some space at the start', async () => {
```
gregtatum, line 29: Does this actually happen with our profiles?

src/routes/publish.js
```diff
@@ -42,29 +45,78 @@ export function publishRoutes() {
...
     // error. However the HTTP request isn't part of the pipeline because we
     // don't want to destroy it. Indeed if it's destroyed, the socket is closed
     // and we can't send a nice error to the caller.
-    const pipeline = util.promisify(Stream.pipeline);
+    const pipeline = promisify(Stream.pipeline);
+    const finished = promisify(Stream.finished);
+
+    // We create 2 interconnected pipelines:
+    //
+    //     request
```
gregtatum, line 62: Thanks for the helpful diagram!

src/routes/publish.js
gregtatum: Works for me, it's a bit unfortunate the streams API is so poorly designed.

src/routes/publish.js
```diff
@@ -42,29 +45,78 @@ export function publishRoutes() {
...
+      // stop flowing once the buffers in the second pipeline are full.
+      lengthChecker.unpipe(gunzipStream);
+      gunzipStream.destroy();
+    });
+
+    // Forward errors between streams in the second pipeline, as well as the
+    // first pipeline through lengthChecker.
+    // Note that errors inside the first pipeline are forwarded thanks to the
+    // `pipeline` thingie.
+    forwardErrors(lengthChecker, gunzipStream, jsonChecker);
```
gregtatum, line 108: Wow, this feels like magic. (To be clear, because the streams API is magical)

test/unit/streams.test.js
julienw: Not when feeding directly from the UI but I'm also thinking of some more "manual" use cases where profiles could be posted (for example from CI). Maybe too much precaution from my side...

src/routes/publish.js
julienw: Forwardum Errorus

PR #89, opened 2020-06-18
Update the load test scripts to generate gzipped json payloads

julienw:
This is what I've been using to test the new server lately.

The files are copy-pasta-ed, sorry for the duplicated changes. Eventually I'll make a package but I don't want to spend time on this now...

### DIFF
--- a/loadtest/publish_short_requests.py
+++ b/loadtest/publish_short_requests.py
@@ -7,6 +7,7 @@ less limited by the available bandwidth.
 import json
 import os
 import base64
+import gzip
 from molotov import scenario, setup, global_setup, teardown, global_teardown
 from molotov import set_var, get_var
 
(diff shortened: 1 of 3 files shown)


loadtest/publish_short_requests.py
```diff
@@ -173,6 +200,12 @@ async def delete(session, jwt_token):
 
 # Each scenario has a weight. Molotov uses it to determine how often the
 # scenario is picked.
+@scenario(1)
+async def publish_and_delete(session):
+    jwt_token = await publish(session=session, data_size=10)
+    await delete(session=session, jwt_token=jwt_token)
```
julienw, line 206: This is a new scenario, it was missing from the "short requests" test file, but already present in the other one.

PR #90, opened 2020-06-18
Make the concatener a Readable and upload to google chunk by chunk

### LINKED ISSUE #83: Change implementation of the Concatenator to make it look like more a stream
The Concatenator works by keeping all chunks in an array, and concatenating them all when the stream is ended. Then we keep and return this result when `transferContents` is called. The result is also cleared when the stream is destroy.

Node v14 now autodestroys stream by default when it ends.
The easy option is to force autoDestroy to true in the constructor.
A  better option could be to output the result at the end like a stream, simply pushing the result of the concatenation. Possibly we would need to use objectMode for this too, so that node handles this result as one object instead of a bug chunk. This would make the object more true to how streams work, and in the end more compatible with this mechanism in the long run.

julienw:
This change halves memory usage during an upload by avoiding a copy, by giving the data to the google stream chunk by chunk instead of all at once (checked using top). This also seems to recover memory better when aborting big requests (but I'll have to double check this).

The first commit is #89, please don't look at it.

This might fix #83 incidentally but I haven't tried yet, I'll do before requesting review.
I already tested with a real profile end-to-end.

There's seem to still have a small leak, but I'm not sure yet. That's why I'm not requesting a review yet.

### DIFF
--- a/src/routes/publish.js
+++ b/src/routes/publish.js
@@ -118,21 +118,26 @@ export function publishRoutes() {
     // Koa which will expose it appropriately to the caller.
     await Promise.all([pipelinePromise, jsonCheckerPromise]);
 
+    // Step 2: Upload the data to Google Cloud Storage.
+    // All data chunks have been stored in the concatener, that we'll read from
+    // to feed the google write stream.
     const storage = gcsStorageCreate(config);
     const hash = hasherTransform.sha1();
-
     const googleStorageStream = storage.getWriteStreamForFile(hash);
 
-    // We can't use Stream.finished here because of a problem with Google's
+    // We don't use `pipeline` because we want to cleanly unpipe and cleanup
+    // when the request is aborted.
+    concatener.pipe(googleStorageStream);
+
+    // We can't use Stream.finished here either because of a problem with Google's
     // library. For more information, you can see
     // https://github.com/googleapis/nodejs-storage/issues/937
     await new Promise((resolve, reject) => {
-      const fullContent = concatener.transferContents();
       googleStorageStream.once('error', reject);
       googleStorageStream.once('finish', resolve);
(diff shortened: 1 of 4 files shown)


src/routes/publish.js
```diff
@@ -118,21 +118,34 @@ export function publishRoutes() {
...
+    // when the request is aborted.
+    concatener.pipe(googleStorageStream);
+
+    ctx.req.on('aborted', () => {
+      log.debug('request-aborted', 'The request has been aborted!');
+      concatener.unpipe(googleStorageStream);
+      const error = new BadRequestError('The request has been aborted.');
+      googleStorageStream.destroy(error);
+      concatener.destroy(error);
+    });
```
julienw, line 138: From my test this doesn't seem to make a big difference. But this does no harm 🤷

src/utils/streams.js
```diff
@@ -124,20 +137,10 @@ export class Concatenator extends Writable {
...
   }
 
-  _final(callback: (error?: Error) => mixed) {
-    this.log.trace('_final()');
-    this.contents = Buffer.concat(this.chunks);
-    this.chunks.length = 0;
-    callback();
-  }
-
   transferContents(): Buffer {
```
julienw, line 140: I kept this one for tests, as it's still useful in that context. But we shouldn't use it for the production code.

I wonder if I should add a check for `NODE_ENV==='test'`... That wouldn't work when running tests obviously but we'd see it when running integration tests. Thoughts?

julienw:
I've thought about this a bit more, this may be just easier to do the work I planned to generate a token before starting to upload, effectively removing the need to concatenate the content. I think i'll try this next week and see if I can get it to work with a small amount of work.

julienw:
Superceded by #91.

PR #91, opened 2020-06-22
Generate a token at the start of the process and use it to upload to Google right away

### LINKED ISSUE #31: Stream profile binary data to cloud storage rather than load it fully in memory first
In #25, the route to upload profiles fully loads it in into memory through the `Concatenator` class. It then computes the hash. This is a bit dangerous, as the server could run out of memory, and be open to DOS attacks.

## Solution 1 - Non-hash file names

This solution involves not computing the hash for the filename, some other scheme would be created for assigning unique names to the profiles.

## Solution 2 - Move the file after upload

This would create a 2 step process for the buckets.

* The file would be stream uploaded to a temporary location.
* After streaming the hash can be computed
* Copy the temp file to the final location
* Delete the temp file
* Report the final hash to the user.

This is a bit riskier since the "move" operation is non-atomic and could fail, plus there can be additional usage charges for the copy operations.

### LINKED ISSUE #33: Possibly rewrite the publish endpoint using rawBody instead of our own Concatenator
In a next PR we'll use a json body handler that uses [rawBody](https://github.com/stream-utils/raw-body) under the hood. Maybe we can try to use it instead of our custom `Contatenator` writable stream. This could help fix #32.

julienw:
This proved easier to do than I thought!

### Assumptions

Here are some of the assumptions of this patch:
* 16 bytes of random data is enough, because we're more interested in avoiding collisions than making the random string unguessable. [According to Wikipedia](https://en.wikipedia.org/wiki/Universally_unique_identifier#Collisions) on v4 UUID (note they have 122 bits of randomness instead of 128 like we do):
  > For example, the number of random version-4 UUIDs which need to be generated in order to have a 50% probability of at least one collision is 2.71 quintillion.
* `base64url` provides a suitable alphabet for the token: upper and lowercase letters, digits, `-` and `_`.
   **Update**: the last commit uses a `base32` alphabet instead of `base64url`.

### Benchmarking

I did a lot of benchmarking uploading a 20MB files (either 1, or several in parallel with at most 50 workers) from my local server to GCS. This isn't perfect because the production server is a lot closer to GCS than my computer, but I believe this is a worse situation because my computer needs to keep the data longer than the real production server would. For raw timings I couldn't see a big difference between this implementation and the original implementation. However the memory usage is vastly improved, as we expected.
I also looked at the effect of buffering the data more than what node does by default (which is 16kb for normal streams, but might be different for HTTP streams), but could see no difference, so I removed this (FTR I tried 64K, 128K, 512K, 1M, 2M).

### Testing strategy

I did local testing with molotov, the unit tests, and end-to-end testing using a local version of the frontend with the constants pointing to the local server and the development storage bucket.

### Corresponding issues
Fixes #31 
Closes #33


src/routes/publish.js
```diff
@@ -9,23 +9,33 @@
...
   forwardErrors,
   GunzipWrapper,
 } from '../utils/streams';
+import { toBase64url } from '../utils/base64url';
 import { PayloadTooLargeError } from '../utils/errors';
 
 const MAX_BODY_LENGTH = 50 * 1024 * 1024; // 50MB
 
+async function generateTokenForProfile(): Promise<string> {
+  const randomBytes = promisify(crypto.randomBytes);
```
julienw, line 30: We use the async version of randomBytes (the one taking a callback) so that it's asynchronous and doesn't block the thread.

src/routes/publish.js
```diff
@@ -118,23 +125,7 @@ export function publishRoutes() {
...
-
-    // We can't use Stream.finished here because of a problem with Google's
-    // library. For more information, you can see
-    // https://github.com/googleapis/nodejs-storage/issues/937
-    await new Promise((resolve, reject) => {
-      const fullContent = concatener.transferContents();
-      googleStorageStream.once('error', reject);
-      googleStorageStream.once('finish', resolve);
-      googleStorageStream.end(fullContent);
-    });
```
julienw, line 134: I started looking at implementing this when I found memory and leaking problems with this. I think that calling `end` with the `fullContent` isn't good, because then we can't make it stop if for example the original request is aborted.

I tried to fix just this in #90, but then thought it was just easier to do the Right Thing.

src/utils/streams.js
```diff
@@ -88,6 +54,8 @@ export class LengthCheckerPassThrough extends Transform {
 /**
  * This writable stream keeps all received chunks until the stream is closed.
  * Then the chunks are concatenated into a unique Buffer that can be retrieved.
+ *
+ * This is used in tests only.
```
julienw, line 58: This is used in the storage mocks and in some streams test. I decided to keep it here for now, but might move it to test-related areas instead later...

test/api/publish.test.js
```diff
@@ -21,21 +20,27 @@ import {
...
+  if (!res.text) {
+    throw new Error(
+      `There was no 'text' property in the response, which shouldn't happen.`
+    );
+  }
+
   const token = res.text;
   const secret = config.jwtSecret;
   const decodedPayload = jwt.verify(token, secret, { algorithms: ['HS256'] });
-  res.text = decodedPayload.profileToken;
```
julienw, line 28: Previously we were replacing the `text` (which is the response's data) with the token so that we could use supertest's `expect` thingie. Now it's a much more classic function :-)

test/api/publish.test.js
```diff
@@ -44,35 +49,20 @@ describe('publishing endpoints', () => {
     return agent.post('/compressed-store').accept(acceptHeader).type('text');
   }
 
-  it('uploads data all the way to google storage when using a content length', async () => {
```
julienw, line 47: This didn't disappear, this is just below. But I extracted quite a big part of this test in a function, that's why github is showing it like this.

test/api/publish.test.js
```diff
@@ -266,21 +254,17 @@ describe('API versioning', () => {
...
       `image/webp,${ACCEPT_VALUE_MIME};version=1`
     );
-    await req
-      .send(BASIC_PAYLOAD)
-      .expect(200)
-      .expect(verifyAndDecodeJwtToken)
-      .expect(BASIC_PAYLOAD_HASH);
+
+    const res = await req.send(BASIC_PAYLOAD).expect(200);
+    verifyAndDecodeJwtToken(res);
```
julienw, line 259: Note I call it as many times as I can, because, as it's random, I want to check characters a lot, to make sure we don't have wrong results sometimes. If we someday have an error this will be intermittent but I want it to fail as often as possible.

src/routes/publish.js
gregtatum: nit: Should we hoist the promisify call out of the function scope so it's only done once?

src/utils/base64url.js
```diff
@@ -0,0 +1,28 @@
...
+// We also skip the padding in this case, removing the ending `=` characters,
+// because that's useless in our use.
+// See https://tools.ietf.org/html/rfc4648#section-5 for more information.
+
+// Useful RegExp expressions to remove or replace the characters from a base64 string.
+const replacePaddingRe = /=+$/; // Replace adjacent '=' characters at the end of string.
+const replacePlusRe = /\+/g; //    Replace all '+' characters anywhere.
+const replaceSlashRe = /\//g; //   Replace all '/' characters anywhere.
+
+export function toBase64url(buffer: Buffer): string {
```
gregtatum, line 18: So what do the "hashes" look like now? It's not clear to me.

julienw:
Ah, that's a good question!
We get things like: `aU4sr5r4tG1xFPRuJbXfqQ` or `_zeTZWj-e7OYqNgDJKzFqQ`.

I was worried about having `_` or `-`. But now, looking at it more, I'm more worried that 1/ upper and lower case characters are different, and 2/ we use characters that can be confused (eg: O et 0, or I, l and 1).

I'm considering switching to a base32 encoding instead of base64. base32 takes roughly 20% more space than base64, so this would still be less than our current hex-based hash.

Advantages of base32 are:
* we can use an alphabet without confusing characters (in input mode, we can normalize the confusing characters to the ones alike in the alphabet)
* we output only lower or upper case (in input mode we can normalize the case), so it's easier to dictate
* I also like Crockford's idea of removing the letter `u` to avoid accidental profanity

I'll do the work as it should be straightforward, and then we can decide.

julienw:
here are examples I get with the base32 alphabet:
```
114361a2dpk9kj2kajph16849c
2b1qp098bh1cary6f7sywcwm80
2qnkktr1ndqz69qbatb00z1ptg
30d0d0dqmsseqysrrnwhncq9f4
30s14jkfeknwy1pxzg86b10sc4
37ymy1akxs12pfsesppjtf1ed0
3sag6cmqe17t4kxvxjky1h3h7g
3sdzegk7d2yqt58m2fypgr50jm
4zs340p6ha14ms54v4j6gdhnqw
5a8qqhjmx3ysfathkwkdaq889m
62myb4e6jjvb1d67ncq4ep4wq4
6v4pk0n6twx4pf79r9keg6zjtm
71h84q6axvew9jv6t0wy3wkvew
73vv30tnff8jq0zzp06g2swjhw
76bkdnnx053t6acn33qwbjhfpg
79s8v4yags424qpvfs58f7r1e8
8rtdx4744m34z2yabz47g79c6r
9a7vvgb0ff1ys23hxfs494ctjw
9tnc51vy1z7m2ss8fyvcq1nyt8
9yps7bjtqp1486e95b83v5apn4
b2b3m0d1mjn6ej0dn3jm0yv3f4
bpv5cxy2w28z72a3e7j3vfbfr4
bvtb3zv3sgjc01pppj9p27p7mc
c7t6kc6r0r2ftpck2591w8k58c
cqv4n1c7aesqj7vsghx8bward0
cqzk33kckc4e6paq350j24xck4
d6n8c0tfkbvr3y1savr8gsgvcw
df1m4m66at9xkfzz77gcs9t2bc
dfy972h0as5v989y4wg4h5pz8c
dn1h4xh39xgn4bd7bsss1mc9j4
dp3xve8chkx4y0qh25eafctwvc
ez2cb3efyw455b8pswstdpzhd8
fbr954py5yyyxx5t7td1z6dgam
fjcdr8e05dgnfavc6vgzb1k6t8
fkzvgy72kz9bhmp2r3yz54h0zg
fq3e060m4z7jwwtsjn2fc0j228
g10e0mf2xvbp3aw2cvfr1zgvgw
ghq2mtjpvpg6exffrkwmyy7hc0
gqeyvsaczn6n2ccxz6twfk76c4
gx26h7q444cw103wg259d4gh7r
hcd0fvdqdrm61h71787gjg2v8w
hwgqeac3m5fzqd4g71spdxm3wr
j600zbx4kc4g3q1bmrpjjatghc
jb94km98j88k7hr2edjmbetcmr
jf16hr3ar1k9mqj7zs2pjnvt0c
jj6m8zpntm5pe82g6pm712jwr4
jywbfqz6ja2ngp38bgwc0vght0
k52f077hw9fpy4c5pvhyxmq4bw
kb2s8ceebcrfdp3gf9f7rpeptr
kwbc1m874jd76sj7jdnt3b2phc
ma1c8rgzzdaqwh7b3vw062t4q8
madhxe11g2z0r620zch5a3a5w8
mq5m0wnxpjr46dj89dz52f32e4
p33jqarsx97cnfbbyvvhkyhev4
pedxh9zw0qwedj2eqv86dgyv64
ph01xnd4jeytmcqmcc5ma6zqqg
pr982grskcayg1fqfddbv6ss5m
sv7fyf35sp791vrkgf717g6nd0
trta7m591vcexjr3s4814220wr
v5bjgv1t3q8f53xsymqrb3jkx8
vn3cb9gkad7vqwhs5t6t031tpg
w0skvt2tzsq7wmd56aaskp3gtm
w1gps40p3qdtjws72epnaby7r0
w5hzrxmxkqjwtnpf49cma4zeqm
x6pceg80jkqfdsjv170p100f04
xp98hy52ym8zza0kyxva4d0d3w
yaxgjfknb6c01dd9dk1hgjrhsm
yr52rkqvw67s8kjnm44pcsjcs8
yr7kk8j873nfrpgc57efvesq6m
yxr9pe8q5fgqexjdk905ycjp5w
zqh5txg3a5vq5qmypeakwwa8gc
ztaeeqs8j1ftbgwmfznz6vaxt8
```

src/utils/base32.js
```diff
@@ -0,0 +1,122 @@
...
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+*/
+// This implementation comes from the project
+// https://github.com/agnoster/base32-js, from this revision:
```
julienw, line 15: This was the library linked from the wikipedia article. I took it because it was self-contained and and the alphabet was easy to change. I slightly audited the code too.

src/utils/base32.js
```diff
@@ -0,0 +1,122 @@
...
+//
+// @flow
+
+// This would be the place to edit if you want a different
+// Base32 implementation
+// Note: We edited the original alphabet to add back the "s" and remove the "u",
+// following Crockford's advice to avoid accidental profanity. Also "u" and "v"
+// can be confused.
+// Otherwise the characters "o", "i", "l" are absent.
+const alphabet = '0123456789abcdefghjkmnpqrstvwxyz';
```
julienw, line 28: nitpicking time!

Some choices that exist in other alphabets:
* 's' is removed because it can be confused with '5' (especially in handwritten form). This was the case in the original implementation, I added it back so that 'u' can be removed.
* 'u' and 'v' can be confused in handwritten form too -- but because I removed the "u" here we're fine about that
* z-base-32 changes the order so that the "easier" characters come more often. I wasn't convinced, but [here is some explanation](http://philzimmermann.com/docs/human-oriented-base-32-encoding.txt) if you want to know more.
* in some other implementations they prefer the letters over the digits, keeping 'o' over '0' or 'i' over '1'.
* I believe 'i' in lowercase is maybe quite recognizable indeed: we could remove 's' and add 'i' back.

src/utils/base32.js
julienw: I imported it so that:
* I can change the alphabet
* it's easier to add the types ;)

src/utils/base32.js
gregtatum: I'm not too opinionated on this.

PR #93, opened 2020-06-25
Add a dependabot configuration file to automate dependencies update

julienw:
I followed the documentation at https://help.github.com/en/github/administering-a-repository/enabling-and-disabling-version-updates.

I want to look at how this works here, where the dependencies are reasonably uptodate, before adding it to the profiler frontend repository.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,25 @@
+version: 2
+updates:
+  # Enable version updates for npm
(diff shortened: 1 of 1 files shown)


canova:
As an optional thing (maybe in the future), it would be good to add performance-admins or specific people as reviewers so we won't miss these PRs: https://help.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates#reviewers

PR #94, opened 2020-06-26
Use 24 bytes instead of 16 for tokens

julienw:
I had a quick discussion with @beurdouche, where he advised that using 24 bytes to generate tokens would bring some more future-proof protection to the uploaded data.

To restate what I put in the comment:
```
  // We're especially interested in avoiding collision, but also interested that
  // we can't easily find a random profile by exhaustively crawling the token
  // space. That's why we use the number of 24 bytes (192
  // bits).
  // * This should be more then enough to avoid collisions, according to the Wikipedia page about UUID:
  //   https://en.wikipedia.org/wiki/Universally_unique_identifier#Collisions
  // * This should be also enough for crawlers, even if 32 bytes is usually
  //   recommended, because we'll expire the data soon, most uploaded profiles
  //   are sanitized by default, and only a fraction have useful information for
  //   an attacker.
```

This token is used to store the published data under a file of this name in Google Cloud Storage, which is then publicly accessible under this name at the URL `https://storage.googleapis.com/profile-store/<TOKEN>`.

In addition to that we realized that the URL shortening by bit.ly is reducing the crawling space too, because it uses 6 (currently) alphanumeric characters on our custom domain share.firefox.dev. Using a custom domain exposes us more too, because we don't share the token space with other uses than the profiler usage. Therefore increasing the randomness might yield no useful improvement for us in real life. A quick check made me think there's no rate limiting on bit.ly end, to access shortened URLs.

24 bytes gives 39 base32 characters, which is about the same size as our old sha1 hex (40 characters). As a bit more context, we recently moved from using the sha1 of the stored data as file name, to generating this token as shown in this patch.

Hey @dveditz, Benjamin told me you might have useful feedback about this change, specifically:
* whether changing from 16 to 24 changes anything in our case
* whether using 32 bytes would be very recommended.

Thanks!


julienw:
Thanks for the review!

dveditz:
How much use does the server get?
* How many devs are using it?
* how many profiles do they save?
* how long are profiles saved for?
* how many devs do we _hope_ use it in the future?

A millions devs saving 10 profiles a week, and we'll assume profiles expire every week, is 10^7 active profiles, in a namespace of ~6x10^57 (up from ~3x10^38). Those are swimming in a lot of empty space! How crowded could it get? Let's assume nearly everyone turns into a developer (~100M daily) saving 25 profiles a week and we save them for 4 weeks. That's ridiculous, but even so it's 10^9 active profiles at any one time. That would take on the order of ~10^48 requests to find a working profile. Even the old number of would take ~10^29 requests. I'd hope that kind of traffic would trigger all kinds of alarms on our server, but even if our server supported 1,000,000 requests a second (ha!) that's still quadrillions of years even in the smaller namespace. Whew, we're safe!

But, bit.ly :-(

bit.ly's namespace is 7 characters (62 possibilities), only ~41 bits or 3.5x10^12 possibilities. Kind of doesn't really matter than the back-end went from 128 bits to 192 bits if our funnel is only 41. If an attacker made a more reasonable 1K requests a second and we go back to a lower assumption of 1M active profiles that would take just under an hour to guarantee they'd find a single profile. At 100 requests a second you could still find a couple a day. Would bit.ly support that kind of volume? They say they have abuse lock-outs which might help here. I'm reasonably confident we don't currently have 1M active profiles, but what's realistic in the future?

It's interesting that the share.firefox.dev/ codes work on the bit.ly/ domain but other bit.ly codes don't work on our domain. If someone had to search actual bit.ly/ links to find our profiles they'd find tons of other random stuff instead along the way. But by searching our specific share domain they know they're only going to find our stuff. Is it possible to hand out raw bit.ly/ links instead? Not as "branded" or memorable, but safer.

Loading the profiles in our website isn't fast enough to do this kind of attack, but you don't have to. If you get the redirect URLs from bit.ly directly (xhr/fetch/curl) you could pull out the random ID and go directly to google storage to download the data. And really, the important thing is how fast does bit.ly fail so you can try the next combination? Seems pretty fast. This seems to be the weak link in the scheme (pun alert). It all comes down to what kind of brute force protection bit.ly uses.

People don't _have_ to use the permalink button, do they? Once you press the Share button and upload them are they saved the same amount of time whether you press "Permalink" or not? But it's so handy...

julienw:
Thanks for sharing! This is quite aligned with my initial thoughts, which is reassuring (but not so much for bitly indeed).

Here are some answers and corrections to the assumptions:
* currently mostly only gecko developers use the tool (I'd say roughly 50 people max), but we plan to replace the old performance panel with this new tool. The old tool sees 3k users daily according to amplitude (which may undermeasure, but this gives an idea). There's hope the new tool will get more users but that won't ever be that crazy :D
* From my very clever maths, I measured in average 450 new profiles every month between this february and now (~2000 since Feb 21).
* for now profiles don't expire, they stay shared forever. We plan to support an expiration system, but by default that would be more probably 6 months in our current plan. We'll still have shared profiles that'll never expire. So I believe this changes your math a bit, by adding 1 or 2 degrees of magnitude. (but the number of users would decrease it too, actually :) )
* bitly generation: currently we generate a shortlink for each published profile. But I had the same idea than you, that we should make it not automatic, and maybe a little bit harder to reach with some warning. The current behavior was tailored to gecko developers where we mostly wanted to share it to a bugzilla bug after uploading. 
* bitly: I wonder if we can ask our contact at bitly to have some more characters (maybe move it up to 10). Then if we don't generate it all the time this could at least reduce the problem.

In the longer term, we may want to stop using bitly altogether, and come up with another scheme.
We could:
* go back to 16 bytes. Maybe less (could 12 be enough randomness ?)? A 16-byte token takes 22 base32 characters, a 12-byte token takes 20 base32 characters.
  * Using 12-byte tokens with the math change to have 10^11 profiles instead of 10^9 gives that we need 10^17 requests to find one profile, which is 3170 years with your 1M request/second assumption. It's not clear to me if this is too low :-) But I'd not take the chance just to reduce by 2 characters only.
* serialize the URL parameters in a much more compact way -> this is the main reason we're using bit.ly, because that can be very long.

The resulting URL would still be a lot longer that what we have now. A few users are copying the URL by hand from a computer to another.

PR #107, opened 2020-06-29
Update dependabot configuration

julienw:
Hey,
I'm updating a bit our new dependabot configuration. Tell me what you think!


.github/dependabot.yml
```diff
@@ -11,6 +11,8 @@ updates:
       - "julienw"
       - "canova"
       - "gregtatum"
+    allow:
+      - dependency-type: "direct"
```
julienw, line 15: I'd like that we get updates only for the direct dependencies, to reduce a bit the load...

According to the doc this has no effect on npm-based projects thought, so maybe this will do nothing and I'll remove it next time.

.github/dependabot.yml
```diff
@@ -23,3 +25,6 @@ updates:
       - "julienw"
       - "canova"
       - "gregtatum"
+    ignore:
+      - dependency-name: "node"
+        versions: [">12"]
```
julienw, line 30: With this we don't track other versions that the one we currently use.

Dockerfile
```diff
@@ -11,9 +11,13 @@
...
 # Setup a container and build the project in it
-# We set up a node 10 running in latest Debian stable called buster, in the
+# We set up a node 12 running in latest Debian stable called buster, in the
 # "slim" flavor because we don't need the big version.
-FROM node:12-buster-slim AS builder
+# By specifying an exact node version, we let dependabot update the version when
+# there's a new one, so that a new docker image is generated. Make sure to use
+# the exact same version for the second part of the Dockerfile.
+# /!\ Make sure to update dependabot.yml when switching to a new major version. /!\
+FROM node:12.18.0-buster-slim AS builder
```
julienw, line 20: I specify a complete version number now. Thanks to dependabot this will force us to update the docker image when there's a new node LTS version. Note there are more update happening, like every day, because in the buster image itself dependencies are updated. So this is a good thing that we get regular updates.

I used 12.18.0 because there _is_ a 12.18.1 version and I want to see if dependabot will update it correctly, in both places in this file.

.github/dependabot.yml
julienw: Note that according to the doc, transitive dependencies will still be checked for security issues.

julienw: Hum actually I think they already track direct dependecies only... i'll remove this.

julienw:
I'll use another approach, so closing this PR :)

PR #108, opened 2020-06-29
Update dependabot configuration to make it closer to our needs

julienw:
This removes docker dependency check (see the rationale below) and bumps up the number of PRs opened by dependabot at once.
Also this upgrades the remaining dependencies manually. I should have done that first :-)


.github/dependabot.yml
```diff
@@ -7,18 +7,9 @@ updates:
     # Check the npm registry for updates every week on monday
     schedule:
       interval: "weekly"
-    reviewers:
-      - "julienw"
-      - "canova"
-      - "gregtatum"
-
-  # Enable version updates for Docker
-  - package-ecosystem: "docker"
```
julienw, line 16: When I configured this, I thought it would be a good idea to update node versions so that we would have newer docker images with the security updates.

But we want to stay on node LTS anyway and upgrade to a new major version only when ready. Updating minor versions can be done by pulling new base images anyway.

So after thinking this more I decided to go with the PR #109 for this. This has the added advantage of pulling security updates for the underlying system even when there's no new node version.

canova: Good call. I agree that this dependabot docker config didn't turn out what we want :/ Thanks for removing it.

.github/dependabot.yml
```diff
@@ -7,18 +7,9 @@ updates:
...
-    # Look for a `Dockerfile` in the `root` directory
-    directory: "/"
-    # Check for updates once a week
-    schedule:
-      interval: "weekly"
+    # We have some linked dependencies (eg: babel) so let's bump up the limit of
+    # open PRs.
+    open-pull-requests-limit: 10
```
canova, line 12: I wish some day dependabot will send only one PR for all the dependency updates combined :)

PR #120, opened 2020-06-30
Add lockfile-lint

julienw:
This is a useful check that we already have in the frontend. I find it even more useful now that we have automated tools changing yarn.lock for us :-)

FTR I tried adding other things we have in the frontend:
* alex: this brings too much violations that I think aren't useful in our case.
* license-check: some of our dependencies fail the test because their "license" property uses an invalid value, even though they use the right license. I filed PR here and there (including in license-check itself to accept these values) so that hopefully we could add it in the future.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "lint": "run-p lint-js lint-python",
     "lint-fix": "run-p lint-fix-js",
     "lint-js": "eslint src bin test",
-    "lint-fix-js": "yarn lint --fix",
+    "lint-fix-js": "yarn lint-js --fix",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -16,7 +16,7 @@
     "lint": "run-p lint-js lint-python",
     "lint-fix": "run-p lint-fix-js",
     "lint-js": "eslint src bin test",
-    "lint-fix-js": "yarn lint --fix",
+    "lint-fix-js": "yarn lint-js --fix",
```
julienw, line 19: This was a left-over mistake that I included in this PR because....

PR #151, opened 2020-09-15
Add the error reason to the FlowExpectedError comments, because newer…

julienw:
… versions of Flow warn about this

This fixes [the Flow error](https://circleci.com/gh/firefox-devtools/profiler-server/2487) in https://github.com/firefox-devtools/profiler-server/pull/150. Starting Flow 0.132, Flow generates warnings to suggest adding the underlying reason of $FlowExpectedError comments. But we configure our Flow so that warnings error our check, because we want that old $FlowExpectedError that don't match anymore be removed. So... we have to add these reasons too :-)


canova (approved):
Looks good to me, thanks!

PR #163, opened 2020-11-09
Upgrade molotov to version 2.0, and add the new way of running all sc…

julienw:
…enarios in one file in the README

Tarek recently released a new version of molotov, where he kindly fixed most issues I found while working with this tool. 
Especially there's a new way of running scenarios, with `--single-run`, which runs each scenario in one specific file once. This is very good for us because this gives us a way to easily use the same files for load testing and integration testing.

In the future it could be a good idea to automatically run these tests on the produced docker image in CI. But that's more work than I'm willing to do now.

### DIFF
--- a/loadtest/README.md
+++ b/loadtest/README.md
@@ -30,6 +30,11 @@ Note: to exit the virtualenv, just run `deactivate`.
 pip install -r requirements.txt
 ```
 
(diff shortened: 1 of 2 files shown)


gregtatum (approved):
Looks good to me! I didn't try it out locally myself.

PR #171, opened 2020-12-08
Deploy Dec 8, 2020

julienw:
* various dependency updates, including security updates
* upgrade the docker version in use


julienw:
I did a manual load test on the dev and stage servers, to check they were similar in terms of load handling. Truth is that the dev server seemed to handle more load.

PR #172, opened 2020-12-08
Upgrade node to the latest versions in the tests

julienw:
This is a very simple patch to get latest version of v12 for tests. The server is already using the latest and greatest version automatically because we're not specifying the minor version, only the major version.


canova (approved):
Looks good to me, thanks!

PR #173, opened 2020-12-09
Upgrade to node v14

julienw:
Manual testing on the server seems to show that it works fine. I still have to test it with a real instance of the profiler UI.
Also one test is failing but I can't find why for now.

### DIFF
--- a/src/utils/streams.js
+++ b/src/utils/streams.js
@@ -63,7 +63,10 @@ export class Concatenator extends Writable {
   contents: Buffer | null = null;
 
   constructor() {
-    super({ decodeStrings: false });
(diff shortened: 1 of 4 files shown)


src/utils/streams.js
```diff
@@ -63,7 +63,10 @@ export class Concatenator extends Writable {
   contents: Buffer | null = null;
 
   constructor() {
-    super({ decodeStrings: false });
```
canova, line 66: Do you know why we don't need the `decodeStrings` anymore?

canova (approved):
Nice! Congrats on making this happen! :tada:

PR #260, opened 2021-10-19
Use the new recommended images from CircleCI and refactor python CI jobs

julienw:
This changes the python image to use the currently recommended image for our python jobs.
See https://circleci.com/developer/images/image/cimg/python for more information.

This also changes some of the jobs so that we don't run nodejs jobs in the python executor. Indeed the python-node image only got node v14 and we'd like to move to node v16 soon (see PR #173).


julienw:
Note: the "lint" job is still expected because it was renamed to "lint-js".

PR #262, opened 2021-10-21
Remove one reference to a bad ansi-regex

julienw:
I followed the result of `yarn audit` locally. This change is a fix for this dependency chain:
nodemon > update-notifier > boxen > ansi-align > string-width > strip-ansi > ansi-regex    

We miss another reference because of a dependency from flow-typed. I filed https://github.com/flow-typed/flow-typed/issues/4151 there and will follow along their progress.


canova (approved):
Thanks for fixing this and following up with the other one!

PR #266, opened 2021-10-27
Upgrade debian image

julienw:
Debian stable bullseye was recently released.

The production image's size is 202MB (up from 197MB). The building image's size is 560MB (up from 535MB -- it would have been 583MB without the change in the python package installation command).


Dockerfile
```diff
@@ -31,7 +31,7 @@ RUN mkdir /app \
 # - python, as we need it to build fast-crc32c when installing with yarn later.
 # - build-essential, that installs all build tools for the same reason.
 RUN apt-get update \
-  && apt-get install -y --no-install-recommends git python build-essential flake8 \
```
canova, line 34: Was `python` package installing both 2 and 3? I thought it was defaulting to one of them, TIL. (Even though it was defaulting to one of them, it means it's unpredictable anyway. It's better to pin to a version istead like you did.)

canova (approved):
Thanks for the upgrade!
I'm not so much worried about the building image's size. In the end it's being discarded anyway. And increase in the production image sounds fine to me.

Dockerfile
julienw: Thanks for your question, finding the answer to that wasn't trivial!

This is subtle and I had to ask around actually: there's no `python` package anymore. However, the package [python-is-python2](https://packages.debian.org/bullseye/python-is-python2) (which depends on `python2` and provides `/usr/bin/python` as a symlink to pyhon2) has the line `Provides: python`, and as a result, it gets installed, and installs python2.

`python-is-python3` exists too, and also provides `/usr/bin/python` as a symlink to `python3`, however it doesn't have the line `Provides: python` in its package control information, so it's not installed when doing `apt-get install python` (which is a good thing: if it had `Provides: python` too, we wouldn't know which one would get installed...).

In our case we don't need python2 (anymore?) so we can be explicit about installing python3.

PR #294, opened 2022-02-24
Update molotov to 2.4 to fix the dependency issues on loadtest

canova:
Previous versions were using a dependency that is dead which is
aiomeasures. It was failing to find that dependency and failing to
install molotov overall. 2.4 removes that dependency.


julienw (approved):
Thanks, I upgraded locally and everything seems to work just as fine!

PR #295, opened 2022-02-25
Make fast-crc32c an optional dependency for M1 chips

canova:
This PR both changes the `fast-crc32c` dependency to be optional and documents this in our README.md file so people who use M1 chips can install the dependencies properly.


julienw (approved):
lgtm, thanks!

PR #298, opened 2022-03-02
Update google-p12-pem to update node-forge dependency

julienw:
This dependency is used from @google-cloud/storage and has 3 security advisories, including https://github.com/advisories/GHSA-8fr3-hfg3-gpgp.


canova (approved):
LGTM!

PR #341, opened 2022-09-07
Migrate the codebase from Flow to Typescript

### LINKED ISSUE #222: Migrate to TypeScript
We want to migrate to TypeScript to follow the trend of the community.

Also we're already not following latest Flow versions because of incompatibilities with our codebase. Moreover Flow announced they're not working for the community.

See https://github.com/firefox-devtools/profiler/issues/2931 for the same bug in the UI, with a lot of good ideas about the process.

canova:
Fixes #222.

It would be better to look at the PR commit by commit. I tried to make the commits as atomic as possible. I'm keeping the flow until the end of the PR and removing it completely at the last commit. To summarize what I've did here:
1. Added the typescript dependency and some type dependencies for third party libraries.
2. Converted the file extensions from js to ts inside the src directory.
3. Used flow-to-ts to do the conversions for these files
4. Manually fixed the other issues.

Then I did the same thing for test directory, added linting and removed the flow.


src/utils/streams.ts
```diff
@@ -242,11 +239,10 @@ export class GunzipWrapper extends Transform {
     this.gunzipStream.once('end', callback);
   }
 
-  _destroy(err: Error | undefined, callback: (error?: Error) => unknown) {
+  _destroy(err: Error | null, callback: (error: Error | null) => unknown) {
     // This line is needed because of the slightly inconsistent
     // signature of callback vs err, and that we can't change.
```
julienw, line 243: you can remove this comment too :-)

src/types/koa.d.ts
```diff
@@ -0,0 +1,7 @@
+import * as router from 'koa';
```
julienw, line 1: nit: I wonder if this file shuold rather be the name of the module, such us `koa-json-body.d.ts`?

src/types/koa.d.ts
```diff
@@ -0,0 +1,7 @@
...
+
+declare module 'koa' {
+  interface Request extends Koa.BaseRequest {
```
julienw, line 4: Can you please add a comment? my understanding is that you're extending an already existing `Request` interface that's already part of koa. I wonder if the `extends Koa.BaseRequest` part is necessary since it's already part of koa's types definition.

__mocks__/@google-cloud/storage.ts
```diff
@@ -46,15 +46,14 @@ class MockBucket {
...
-  ...
+  metadata: Object;
 };
 
 class MockFile {
   bucket: MockBucket;
   name: string;
   contents: Buffer;
-  metadata: Metadata;
+  metadata: any;
```
julienw, line 55: Is this expected? I see the `Metadata` type is still defined above.

test/api/dockerflow.test.ts
```diff
@@ -22,7 +22,9 @@ describe('dockerflow endpoints', () => {
 
   function setupForHeartbeat({
     failBitly,
-  }: $Shape<{| +failBitly: boolean |}> = {}) {
+  }: Partial<{
+    failBitly: boolean;
+  }> = {}) {
     let nockScope = nock(BITLY_HOSTNAME, {
```
julienw: I don't see types for `nock`, is that expected?

update: I see that nock comes with its own types.

test/api/dockerflow.test.ts
```diff
@@ -131,11 +133,13 @@ describe('dockerflow endpoints', () => {
     });
 
     it('answers to the version endpoint when the file is present', async () => {
+      // eslint-disable-next-line @typescript-eslint/no-var-requires
```
julienw, line 153: I wonder if we could remove this rule for tests only. What do you think?

test/api/utils/check-security-headers.ts
```diff
@@ -2,10 +2,10 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 // @flow
-import { typeof Test } from 'supertest';
+import { Test } from 'supertest';
 
 export function checkSecurityHeaders(request: Test) {
-  if (request.called) {
+  if ((request as any).called) {
```
julienw, line 8: I'm a bit surprised about this, are the types for supertest not correct? Should we correct them by overloading the type instead of using any?

test/api/cspreport.test.ts
```diff
@@ -10,7 +10,9 @@ import { checkSecurityHeaders } from './utils/check-security-headers';
 
 describe('cspreport endpoint', () => {
   function getPreconfiguredRequest() {
-    jest.spyOn(process.stdout, 'write').mockImplementation(() => {});
+    jest
+      .spyOn(process.stdout, 'write')
+      .mockImplementation((_: string | Uint8Array) => true);
```
julienw, line 14: ah, crap :) maybe we should do that in setup or something? I don't know...

test/api/dockerflow.test.ts
```diff
@@ -72,17 +72,26 @@ describe('dockerflow endpoints', () => {
     });
 
     it('answers an error to the heartbeat if the bucket does not exist', async () => {
-      jest
-        .spyOn(Bucket.prototype, 'exists')
+      // We can't use jest.spyOn here because of a typescript issue with jest.
+      // When we use jest.spyOn, the typescript compiler complains that the
+      // mock return value must be void. This is happening because apparently
+      // typescript doesn't understand the overloaded Bucket.prototype.exists method.
```
julienw, line 77: Isn't the problem that `exists` isn't in the types for the gogole storage library? I can see it in the js files but not the type files. Could we improve the types locally?

.circleci/config.yml
```diff
@@ -47,7 +47,7 @@ workflows:
       - lint-js
       - build
       - yarn_lock
-      - flow
+      - ts
```
julienw, line 50: nit: can we call the circleci job `typescript` so that it's more explicit?

bin/post-rewrite.js
```diff
@@ -1,5 +1,3 @@
-// @noflow
```
julienw, line 1: should we typescript check this file?

bin/check-warn-yarn-changed.js
```diff
@@ -1,4 +1,3 @@
-// @noflow
```
julienw, line 1: what do you think about type checking this file?

package.json
```diff
@@ -21,19 +16,21 @@
...
     "serve": "node dist/index.js",
-    "start": "nodemon --exec babel-node -w .env -w src src/index.js",
+    "start": "nodemon --exec babel-node --extensions \".ts\" -w .env -w src src/index.ts",
     "start:prod": "run-s build:clean build serve",
     "test": "jest",
     "test-python": "./tools/test_decode_jwt_payload.py",
     "test-lockfile": "lockfile-lint --path yarn.lock --allowed-hosts yarn --validate-https",
-    "test-all": "run-p flow lint test test-python test-lockfile license-check"
+    "test-all": "run-p ts lint test test-python test-lockfile license-check",
+    "ts": "tsc",
```
julienw, line 32: small nit, I wonder about naming the script `tsc`, but I'm not sure what the convention in other projects is.

julienw (changes requested):
By grepping on `Flow` (with the capital letter) I see a few occurrences that should likely be updated or removed.

I believe you ran yarn-deduplicate already, thank you!

The call in https://github.com/firefox-devtools/profiler-server/blob/master/test/integration/test-publish-and-delete.sh#L25 needs to be updated. I know it's not in CI because it's not ready for that, but you can try running the script locally. I also tried to fix it locally but it's not obvious: when replacing `js` with `ts` I get errors about the fact we can't use import, then I used `type: module` in package.json but that doesn't seem the right way either according to some posts on stackoverflow.. Anyway I'll let you look at it closer ;-)

Can you please add your changes to extra commits, so that I can review them easily? Thanks :-)

test/api/dockerflow.test.ts
canova: No, google storage library is a typescript project so it includes types already. `exists` is in here: 
https://github.com/googleapis/nodejs-storage/blob/21f173eb17d4216e2b42ffdd1ed0104aeda7cf13/src/nodejs-common/service-object.ts#L323-L353
This is from the `ServiceObject` base class. [`Bucket` extends from this one: ](https://github.com/googleapis/nodejs-storage/blob/21f173eb17d4216e2b42ffdd1ed0104aeda7cf13/src/bucket.ts#L684)
The issue is that Typescript doesn't work well with the overloaded exists functions above. There is a github issue also in the commit message if you are interested.

canova (commented):
Thanks for the review Julien. Addressed your comments. It's ready for another review now. I've added 5 more commits and didn't change the other ones. You can start looking at them starting from ["Remove skipLibCheck field and fix the errors"](https://github.com/firefox-devtools/profiler-server/pull/341/commits/07b4d304322cafb2d874ccf50d381cac0bb3da44).


> By grepping on `Flow` (with the capital letter) I see a few occurrences that should likely be updated or removed.

Looks like I missed them. Removed them now, thanks!

> I believe you ran yarn-deduplicate already, thank you!

Yep :)

> The call in https://github.com/firefox-devtools/profiler-server/blob/master/test/integration/test-publish-and-delete.sh#L25 needs to be updated. I know it's not in CI because it's not ready for that, but you can try running the script locally. I also tried to fix it locally but it's not obvious: when replacing `js` with `ts` I get errors about the fact we can't use import, then I used `type: module` in package.json but that doesn't seem the right way either according to some posts on stackoverflow.. Anyway I'll let you look at it closer ;-)

Oh I see. I didn't see this test for some reason. I resolved it now. The main problem is that babel-node needs an `--extensions ".ts"` argument. We don't need to add `type: module` to package.json file.

src/log.ts
```diff
@@ -16,7 +16,7 @@ type LowerCasedLogLevel =
   | 'error' //     This is for non-fatal errors.
   | 'critical'; // This is for fatal errors.
 
-function toValidLogLevel(logLevel: ?string) {
+function toValidLogLevel(logLevel?: string) {
```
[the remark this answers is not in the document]
julienw: I see `void` means either `null` or `undefined`, TIL :-)

src/types/koa.d.ts
julienw: ok, thanks for the explanation! My rationale was that this change is to accomodate the code from the module `koa-json-body`.  I'm not sure what the best practices in the TS world are in this regard, but anyway this sounds good to me :) Nothing set in stone anyway.

__mocks__/@google-cloud/storage.ts
julienw: yeah, the new solution looks better to me :-)

package.json
julienw: let's keep it this way, we can always change it later if we're bored :-)

tsconfig.json
```diff
@@ -0,0 +1,25 @@
...
+    // Allow esnext syntax. Otherwise the default is ES5 only.
+    "target": "esnext",
+    "lib": [
+      "esnext",
+    ],
+    "noEmit": true,
+    // Emit additional JavaScript to ease support for importing CommonJS modules.
+    // This enables 'allowSyntheticDefaultImports' for type compatibility.
+    "esModuleInterop": true,
+    // Skip type checking all .d.ts files.
```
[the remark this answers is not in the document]
julienw: Thanks for fixing those!
Isn't it a bit werd that typescript didn't tell you anything about these errors when it was parsing these? Was it ignoring these files at all?

julienw (approved):
lgtm, thanks a lot for this work!

PR #344, opened 2022-09-26
Remove some dependencies

julienw:
This removes 2 dependencies:
1. `@babel/plugin-transform-react-jsx` was needed for the eslint flowtype plugin. Because we don't use Flow anymore, this plugin has been removed, but we forgot the babel plugin.
2. `dts-gen` has some old dependencies, and we have a security advisory because of this, as well as a lot of outdated packages downloaded (see the yarn.lock changes). I suggest we remove it from our dependencies and instead install it on demand only when needed.

Tell me what you think :-)

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "test-lockfile": "lockfile-lint --path yarn.lock --allowed-hosts yarn --validate-https",
     "test-all": "run-p ts lint test test-python test-lockfile license-check",
     "ts": "tsc",
-    "ts-generate-types": "dts-gen -m"
+    "ts-generate-types": "npx --yes dts-gen -m"
   },
   "devDependencies": {
     "@babel/cli": "^7.18.10",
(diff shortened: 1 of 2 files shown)


canova (approved):
> 1. `@babel/plugin-transform-react-jsx` was needed for the eslint flowtype plugin. Because we don't use Flow anymore, this plugin has been removed, but we forgot the babel plugin.

Thanks for catching it!


> 2. `dts-gen` has some old dependencies, and we have a security advisory because of this, as well as a lot of outdated packages downloaded (see the yarn.lock changes). I suggest we remove it from our dependencies and instead install it on demand only when needed.

Hm interesting. I wasn't very happy with this dts-gen library anyway but it was the best I could find at that time. I was surprised that there wasn't a better supported and more user friendly one like flow-typed. I think it's really not that much needed anyway so it's a good idea to install it when we need it instead of keeping it in the package.json. Thanks!

PR #367, opened 2023-01-01
Implemented the redirection using router.redirect

### LINKED ISSUE #165: Implement the redirection using router.redirect
We can replace this code:
https://github.com/firefox-devtools/profiler-server/blob/e630a20c9ebd0d29d737e4237e71bd01e38027f8/src/routes/root.js#L31-L34

with [router.redirect()](https://github.com/koajs/router/blob/master/API.md#module_koa-router--Router+redirect).

In this issue we'll also want to make sure this route is covered by a test. If a test is already there there's no need to add one though.

AvinashDhillor:
Fixes #165 
I had made the code changes, No need to add a test for this route as it was already there.


julienw (approved):
Thanks for the cleanup!

PR #380, opened 2023-01-30
Update molotov to v2.6

julienw:
The newer version of molotov supports Python 3.10 and 3.11, and drops support for Python 3.6. This should be good for us, python 3.6 is very old. See for example [the python3 packages on Debian distributions](https://packages.debian.org/search?keywords=python3), even Debian "oldstable" is using Python 3.7.

This version has an updated console output, which doesn't work in docker containers, that's why I changed this a bit too.


canova (approved):
LGTM!

PR #386, opened 2023-02-14
Update node to v18

julienw:
I didn't check the performance of node v18 compared to node v16. I'll do that before merging to production.
Otherwise everything seems to work just fine.


canova (approved):
LGTM!

PR #430, opened 2023-08-23
Run prettier independently than eslint

julienw:
The way we were configuring prettier was useful when prettier wasn't integrated with IDEs much at the time. Nowadays the recommended approach is to run both tools separately.

It's probably better to look at the commits separately.


canova (approved):
Looks good to me, thanks!

julienw:
I forgot to include the yarn.lock changes, now it's here.
Thanks for the review!

PR #443, opened 2023-11-08
Increase the upload limit to 150 MB and update the error message in logs

julienw:


canova (approved):
LGTM. Commit message still says 100MB though, could you update that as well?

PR #458, opened 2024-01-03
Update debian to bookworm

julienw:
This updates to the latest Debian stable (out for ~6 months now).

I built locally both docker files, and also ran both docker images locally.

After merging to the master branch, I'll run the load test on the deployed docker, as a sanity check. I don't expect a big change.


canova (approved):
Looks safe to me, thanks!

PR #346, opened 2022-10-04
Use the default docker version for setup_remote_docker now

julienw:
This has been recommended by circleci in a mail I received. I couldn't find another reference online other than https://discuss.circleci.com/t/default-docker-version-for-remote-docker-jobs-changing-on-september-6/45157 :/

Here is a copy of this mail:
image

Possible risks: our docker image possibly wouldn't work with the server we use. I'll reach out to our IT to get more information.

PR #466, opened 2024-02-23
Update the images used in CircleCI in preparation for the deprecation

julienw:
Updating various images because of the deprecation notices on CircleCI.
I believe only the change to setup_remote_docker is necessary (see https://discuss.circleci.com/t/remote-docker-image-deprecations-and-eol-for-2024/50176), but it's a good opportunity to update the others.

We can't test whether it works (the update is on the `docker-publish` task), but we'll be able to adjust on the main branch before deploying to prod. Though it's good to mention that the change I made on the `docker-publish` task aligns it with the setup for the `docker` task (to use the default docker configuration in `setup_remote_docker`). This was changed in #346 but looks like `docker-publish` was forgotten back at the time.


canova (approved):
LGTM.

PR #485, opened 2024-05-16
Use Cache-Control: private on uploaded profile data so that it's not …

julienw:
…cached in proxies

I tested locally with my bucket profile-store-julien-dev that this fixes the reported problem. I also checked that the data file is still cached locally in the browser after a reload.


canova (approved):
Oh nice, I wasn't aware of `cache-control: private`. Looks good to me.

PR #538, opened 2025-01-31
Update to node v22

julienw:
Everything seems to work just fine after the update.

There is a warning when running the server:
```
(node:299536) [DEP0048] DeprecationWarning: The `util.isError` API is deprecated. Please use `ObjectPrototypeToString(e) === "[object Error]" || e instanceof Error` instead.
```
I traced that down to https://github.com/seanmonstar/intel/pull/58 (pull request opened by @soulgalore well-known by us :-) ), where `intel` is a dependency for https://github.com/mozilla/mozlog. Both these libraries do not appear well maintained at the moment. We could maintain mozlog ourselves and possibly fork `intel`, but that's not really needed at the moment because everything still works, and I don't imagine nodejs removing that API really soon.

I'll run some loadtesting on the staging server after the merge.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2.1
 executors:
   node:
     docker:
-      - image: cimg/node:18.19
+      - image: cimg/node:22.14
     environment:
       IMAGE_NAME: mozilla/profiler-server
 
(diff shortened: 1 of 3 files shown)


canova (approved):
LGTM!

PR #547, opened 2025-02-20
Update the load testing script to make it shorter

julienw:
Last time we looked at it with @canova, I remembered we thought that the 5 minutes (an arbitrary duration I picked when I initially worked on that) were too long. So I reduced this to 2:30 minutes instead. This makes the duration for the entire script about 15 minutes instead of 30 minutes.

Please tell me what you think @canova !

### DIFF
--- a/tools/load-testing-before-deployment.sh
+++ b/tools/load-testing-before-deployment.sh
@@ -39,16 +39,16 @@ run_tests() (
 
   echo ">>> Ramp up"
   echo "Do not pay attention to the results for this run."
-  molotov -w 200 -d 300 --ramp-up 60 publish_short_requests.py
+  molotov -w 200 -d 150 --ramp-up 60 publish_short_requests.py
 
-  echo ">>> Run a load test operation: 200 clients during 5 minutes (1/3)"
-  molotov -w 200 -d 300 publish_short_requests.py
+  echo ">>> Run a load test operation: 200 clients during 2.5 minutes (1/3)"
+  molotov -w 200 -d 150 publish_short_requests.py
 
-  echo ">>> Run a load test operation: 200 clients during 5 minutes (2/3)"
-  molotov -w 200 -d 300 publish_short_requests.py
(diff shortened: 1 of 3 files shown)


tools/load-testing-before-deployment.sh
```diff
@@ -39,16 +39,16 @@ run_tests() (
...
-  molotov -w 200 -d 300 publish_short_requests.py
+  molotov -w 200 -d 150 publish_short_requests.py
 
   echo ">>> Run a load test operation: 200 clients during 5 minutes (2/3)"
-  molotov -w 200 -d 300 publish_short_requests.py
+  molotov -w 200 -d 150 publish_short_requests.py
 
   echo ">>> Run a load test operation: 200 clients during 5 minutes (3/3)"
-  molotov -w 200 -d 300 publish_short_requests.py
+  molotov -w 200 -d 150 publish_short_requests.py
```
julienw, line 51: Note: I kept all 3 runs because we might still have some hiccups on one of them sometimes. I actually got one today when running them. Running 3 times makes it clearer what is a hiccup and what is normal (usually it's 2 vs 1).

tools/load-testing-before-deployment.sh
```diff
@@ -39,16 +39,16 @@ run_tests() (
 
   echo ">>> Ramp up"
   echo "Do not pay attention to the results for this run."
-  molotov -w 200 -d 300 --ramp-up 60 publish_short_requests.py
+  molotov -w 200 -d 150 --ramp-up 60 publish_short_requests.py
 
   echo ">>> Run a load test operation: 200 clients during 5 minutes (1/3)"
```
canova, line 44: Can we also update the comments that are referencing "5 minutes" here and below?

canova (approved):
Indeed 30 minutes is too long. This looks good to me, thanks!

tools/load-testing-before-deployment.sh
julienw: Oh right, I wanted to do it and forgot !! Thanks for catching it

PR #552, opened 2025-03-04
Add the git directory to dockerignore

julienw:


julienw:
AH `generate-version-file` needs the git repository at the moment, let's see if I can fix that.

julienw:
Ah it needs git to get some information for the version file too, see https://api.profiler.firefox.com/__version__
So I believe we'll keep it working like before. Closing!

PR #573, opened 2025-06-02
Updated circleci config to push to gar

bwells-moz:


scripts/store_git_info.sh
```diff
@@ -0,0 +1,3 @@
+# version.json generated from ./bin/generate-version-file.js
+git log -n 1 --no-color --pretty='%s' > ./commit-summary.txt
+git log -n 1 --no-color --pretty=medium > ./commit-description.txt
```
julienw, line 3: is it needed?
If these files need to end up in the docker image, maybe we can integrate this in ./bin/generate-version-file.js too, just tell me.

`./bin/generate-version-file.js` is called in the Dockerfile: https://github.com/firefox-devtools/profiler-server/blob/12ff914de27f9bd571a9c583f8295c093b77ea6d/Dockerfile#L74-L75

bwells-moz: I've removed the shell command

.circleci/config.yml
```diff
@@ -187,3 +196,43 @@ jobs:
...
+    steps:
+      - checkout
+      - run:
+          name: Prepare environment variables for OIDC authentication
+          command: |
+            echo 'export GOOGLE_PROJECT_ID="moz-fx-profiler-prod"' >> "$BASH_ENV"
+            echo "export OIDC_WIP_ID=$GCPV2_WORKLOAD_IDENTITY_POOL_ID" >> "$BASH_ENV"
+            echo "export OIDC_WIP_PROVIDER_ID=$GCPV2_CIRCLECI_WORKLOAD_IDENTITY_PROVIDER" >> "$BASH_ENV"
+            echo "export GOOGLE_PROJECT_NUMBER=$GCPV2_WORKLOAD_IDENTITY_POOL_PROJECT_NUMBER" >> "$BASH_ENV"
+            echo "export OIDC_SERVICE_ACCOUNT_EMAIL=$GCP_SERVICE_ACCOUNT_EMAIL" >> "$BASH_ENV"
```
julienw, line 220: Can you please precise where these `GCPV2_*` variables come from?

.circleci/config.yml
```diff
@@ -187,3 +196,43 @@ jobs:
...
+    parameters:
+      build_tag:
+        type: string
+        default: build
+      image:
+        type: string
+        default: firefoxprofiler
+      registry-url:
+        type: string
+        default: us-docker.pkg.dev
```
julienw, line 210: Are we gonna change these parameters sometimes?

.circleci/config.yml
```diff
@@ -60,7 +62,14 @@ workflows:
               only:
                 - master
                 - production
-
+      - deploy-gar:
```
julienw, line 65: we probably need also `requires` like above, to make sure the docker image has been produced and cached

.circleci/config.yml
```diff
@@ -60,7 +62,14 @@ workflows:
...
+          context:
+            - gcpv2-workload-identity
+          filters:
+            branches:
+              only:
+                - master
+                - production
```
julienw, line 72: you can comment the filter for now, so that the job will be run in the PR too. We'll remove the comments for the filter before landing!

.circleci/config.yml
```diff
@@ -187,3 +196,43 @@ jobs:
...
+            echo 'export GCP_GAR_REPO="profiler-prod"' >> "$BASH_ENV"
+            echo "export GAR_IMAGE=\"<<parameters.registry-url>>/${GOOGLE_PROJECT_ID}/${GCP_GAR_REPO}/<<parameters.image>>\"" >> $BASH_ENV
+      - gcp-cli/setup:
+          use_oidc: true
+      - run:
+          name: Deploy to Google Artifact Registry
+          command: |
+            gcloud auth configure-docker us-docker.pkg.dev --quiet
+                    
+            source $BASH_ENV
```
julienw, line 230: I believe the `source` command isn't needed, I think it's sourced automatically at the beginning of each step.

.circleci/config.yml
```diff
@@ -187,3 +196,43 @@ jobs:
...
+      - run:
+          name: Prepare environment variables for OIDC authentication
+          command: |
+            echo 'export GOOGLE_PROJECT_ID="moz-fx-profiler-prod"' >> "$BASH_ENV"
+            echo "export OIDC_WIP_ID=$GCPV2_WORKLOAD_IDENTITY_POOL_ID" >> "$BASH_ENV"
+            echo "export OIDC_WIP_PROVIDER_ID=$GCPV2_CIRCLECI_WORKLOAD_IDENTITY_PROVIDER" >> "$BASH_ENV"
+            echo "export GOOGLE_PROJECT_NUMBER=$GCPV2_WORKLOAD_IDENTITY_POOL_PROJECT_NUMBER" >> "$BASH_ENV"
+            echo "export OIDC_SERVICE_ACCOUNT_EMAIL=$GCP_SERVICE_ACCOUNT_EMAIL" >> "$BASH_ENV"
+            echo 'export GCP_GAR_REPO="profiler-prod"' >> "$BASH_ENV"
+            echo "export GAR_IMAGE=\"<<parameters.registry-url>>/${GOOGLE_PROJECT_ID}/${GCP_GAR_REPO}/<<parameters.image>>\"" >> $BASH_ENV
```
julienw, line 222: For more simplicity, and if they are needed, I would create these env variables directly in the step below, like we do for `IMAGE_VERSION_TAG` `IMAGE_LATEST_TAG`.

For the hardcoded ones, it could be easier to add an `environment` property to the job or step like described in parts 2 and 3 in https://circleci.com/docs/set-environment-variable/.

Using the commands with `$BASH_ENV` are useful if:
* we want some interpolation
* we want these variables present for all the following steps

But I'm not opposed keeping them here if you prefer, buy only if we need them.

julienw (commented):
If using the gcp-gcr orb, we should use [gcr-auth](https://circleci.com/developer/orbs/orb/circleci/gcp-gcr#commands-gcr-auth) and [push-image](https://circleci.com/developer/orbs/orb/circleci/gcp-gcr#commands-push-image) from that orb directly.

The gcp-cli is probably not needed!

.circleci/config.yml
bwells-moz: Sounds good

.circleci/config.yml
bwells-moz: My mistake, the context wasn't setup for the firefox-devtools org. Opened a pr on gpa https://github.com/mozilla-it/global-platform-admin/pull/3312

PR #577, opened 2025-06-13
Automatically request reviews for dependency updates

julienw:
Similar to https://github.com/firefox-devtools/profiler/pull/5490


canova (approved):
LGTM!

PR #614, opened 2025-11-07
[WIP] Migrate from CirleCI to GitHub Actions

canova:
This PR is the attempt to migrate from CircleCI to GitHub Actions.

The most important thing is the docker publish flow. We should be very careful not to break it. One change I made was to change the image version tag.
Previously it was `IMAGE_VERSION_TAG="0.0.${CIRCLE_BUILD_NUM}"`. but I've bumped the tag to `IMAGE_VERSION_TAG="0.1.${{ github.run_number }}"` to make sure they don't overlap.

It needs more testing related to how we generate docker images.


canova:
Huh weird, it didn't happen in the profiler repo, but I had to create a PR from this repo and not from my fork to be able to trigger the CI. Internet mentions that this is a security feature, but it should've work since I'm a maintainer.

PR #696, opened 2026-07-06
Accept JsonSlabs files in publish endpoint

mstange:
In the profiler front-end we want to start uploading profiles in the [JsonSlabs (JSLB) binary container format](https://github.com/mstange/json-slabs), so that we don't have to spend time encoding long arrays of numbers to JSON strings.

The server currently rejects such files because it scans for an opening curly brace. This PR adapts the validation code to allow JSLB files.


src/utils/streams.ts
```diff
@@ -148,40 +164,56 @@ export class CheapJsonChecker extends Writable {
...
-        case 'error':
-          this.log.verbose(
-            'json-error',
-            'This stream does not look like a JSON.'
-          );
-          callback(new BadRequestError(this.errorMessage));
-          return;
-        default:
-          throw assertExhaustiveCheck(checkResult);
+    if (!this.jslbRuledOut) {
```
canova, line 172: Hm, what happens if the object size of a JSON profile is lower than the jslb header size? Do we reject that still. In reality I think it's not a problem as we don't usually have profiles that are shorter than the jslb size but it might be still nice to fix that :)

I think for example sending this will fail even though it's a valid json: `'{}'`

test/unit/streams.test.ts
```diff
@@ -44,34 +47,38 @@ describe('CheapJsonChecker', () => {
...
+    const checker = new CheapContentChecker();
     const input = new PassThrough();
     const endPromise = once(checker, END_EVENT_NAME);
     input.write(fixture);
     input.end();
     await expect(pipeline(input, checker)).rejects.toThrow(
-      "The payload isn't a JSON object."
+      "The payload isn't a JSON object or a JSLB file."
+    );
+    expect(endPromise).rejects.toThrow(
```
canova: An these looks like a pre-existing issues, but it looks like we don't have await in from of these `expect`s like the line above. Would you mind await'ing them here while you're touching?

(we also have 2 more of similar checks below)

test/unit/streams.test.ts
```diff
@@ -18,11 +18,14 @@ import {
 const pipeline = util.promisify(Stream.pipeline);
 const nextTick = util.promisify(process.nextTick);
 const END_EVENT_NAME = 'profiler:checkEnded';
+const JSLB_MAGIC = Buffer.from([
+  0xdc, 0xdf, 0x4a, 0x53, 0x4c, 0x42, 0x01, 0x00,
+]);
```
canova: Not for this PR, but would it make sense to have magic being exported from the package directly? I couldn't see it while checking. I don't know if it makes sense from the API of the slabs point of view, but it could help with testing.

canova (changes requested):
Looks reasonable to me! I have 1 small question regarding very short jsons, but happy to go either way. Let me know what you think!

test/unit/streams.test.ts
mstange: Hmm, I suppose we could. Alternatively we could use the builder API in the test to make an actual file.

canova: Ah yeah, that might work better.

test/unit/streams.test.ts
mstange: fixed

src/utils/streams.ts
mstange: Good catch, fixed and added a test.

canova (approved):
Looks good to me, thanks!

It's been a while since we did a deploy for the profiler-server though. It might make sense to do a deployment before this so we make sure that we don't regress anything in the meantime. I'll try to deploy today or tomorrow.

